PR #684, opened 2014-08-21
EXT: Add current user into http response header for logging purposes

flopma:
This can be useful when running Nexus behind a http proxy. This proxy can log each requests with its associated user.


flopma:
Hello, any plans to integrate this feature ? Thx

PR #911, opened 2014-12-10
[NEXUS-7785] allow user-id set by shiro to be used in request-log

jdillon:
https://issues.sonatype.org/browse/NEXUS-7785

A bit hackish, but couldn't find another simpler solution due to the wrapping nature of shiro implementation and the request-log reliance on the underlying HttpServletRequest.getRemoteUser() which is alway null in the context of shiro authenticated request (since its wrapping to adorn those details and the wrapper doesn't live beyond the request context below the SecurityFilter and thus isn't available to the RequestLogHandler).


flopma:
Hi,

I don't understand by reading Shiro docs why you use the SecurityFilter to get the subject for logging.

Getting it from NexusHttpAuthenticationFilter is not enough (parent of NexusAuthenticationFilter, NexusApiKeyAuthenticationFilter and ContentAuthenticationFilter ?

PR #931, opened 2014-12-24
[NEXUS-7812] Drop init message, add startup message and move down

jdillon:
https://issues.sonatype.org/browse/NEXUS-7812


jdillon:
will leave this asis for now, will revisit later as time permits to put a message after jetty has started.

PR #947, opened 2015-01-09
NEXUS-7802: Revisit logging, Take 2

cstamas:
Logging revisited. Introduced "logging name" of task based
on it's configuration to make it more user friendly than
UUID (actually two UUIDs, one for NX Task and one for QZ Job!)
Also, simplified logging (ie. using "Task" instead of "NX Task", etc).

UUID should NOT appear in logs if loggers at INFO level. Still, at DEBUG
and TRACE they ARE logged, to be able to identify them in case of DB dump
analysis.

Also, removed one unused method from QuartzTaskExecutorSPI.

Issue
https://issues.sonatype.org/browse/NEXUS-7802

CI
http://bamboo.s/browse/NX-OSSF476


jdillon:
#goodenough

+1

PR #950, opened 2015-01-09
NEXUS-7835: Tasks to backup and update/recreate npm metadata

cstamas:
Issue covers multiple use cases, several implemented:
* backup of npm db (with possibility of "offline" restore, by copying backup files back)
* update/recreate npm metadata (with data loss in case of DB built from scratch)

Issue
https://issues.sonatype.org/browse/NEXUS-7835

CI
http://bamboo.s/browse/NX-OSSF479


pom.xml
```diff
@@ -575,6 +575,7 @@
               <exclude>**/nexus-csrfguard/**/org/sonatype/nexus/csrfguard/CsrfGuardFilter.java</exclude>
               <exclude>**/nexus-csrfguard/**/org/sonatype/nexus/csrfguard/CsrfGuardServlet.java</exclude>
               <exclude>**/nexus-ldap-common/**/org/sonatype/security/ldap/LdapEncoder.java</exclude>
+              <exclude>**/nexus-core/**/com/bolyuba/nexus/plugin/npm/internal/Version.java</exclude>
```
jdillon: should be **/nexus-npm-repository-plugin/**/com/bolyuba/nexus/plugin/npm/internal/Version.java

cstamas: True, fixed. But still, I did try:  does not matter what you have in there, as even with `nexus-core` it worked... the Version class got ignore just fine.

mcculls: Probably because there's already an exclusion for NPM on line 585 :)

cstamas: Oh, true! Whole npm plugin is excluded.

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+
+    // ie. all processPackageJson failed?
+    if (((Map<String, Object>) rootJson.get("versions")).isEmpty()) {
+      return;
+    }
+
+    final Map<String, Object> lastPackage = packages.get(packages.lastKey());
+    final String lastVersion = String.valueOf(lastPackage.get("version"));
+    final Map<String, Object> distTags = (Map<String, Object>) rootJson.get("dist-tags");
+    distTags.remove(lastVersion);
```
mcculls: I would have assumed the lastVersion would still be kept as a tag, or does npm only keep tags for non-latest versions?

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+    distTags.put("latest", lastVersion);
+
+    final PackageRoot packageRoot = new PackageRoot(context.getRepository().getId(), rootJson);
+    hostedMetadataService.consumePackageRoot(packageRoot);
+    processedPackages = processedPackages + 1;
+    processedPackageVersions = processedPackageVersions + packages.size();
```
mcculls: would it make more sense to name the `packages` variable as `versions` or `packageVersions`?

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+
+    final Map<String, Object> lastPackage = packages.get(packages.lastKey());
+    final String lastVersion = String.valueOf(lastPackage.get("version"));
+    final Map<String, Object> distTags = (Map<String, Object>) rootJson.get("dist-tags");
+    distTags.remove(lastVersion);
+    distTags.put("latest", lastVersion);
+
+    final PackageRoot packageRoot = new PackageRoot(context.getRepository().getId(), rootJson);
+    hostedMetadataService.consumePackageRoot(packageRoot);
+    processedPackages = processedPackages + 1;
```
mcculls: is this really the number of processed packages, or the number of processed package roots? (or is there no difference in npm?)

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+    TarArchiveInputStream tarArchiveInputStream = new TarArchiveInputStream(file.getInputStream());
+    TarArchiveEntry tarEntry = tarArchiveInputStream.getNextTarEntry();
+    while (tarEntry != null && !"package.json".equals(tarEntry.getName())) {
+      tarEntry = tarArchiveInputStream.getNextTarEntry();
+    }
+    if (tarEntry == null || !"package.json".equals(tarEntry.getName()) || !tarEntry.isFile()) {
+      return null;
+    }
+    // TODO: some sanity check of package.json size? If so, for how much (as they grow as package gets "older")
+    checkArgument(tarEntry.getSize() < 100000, "Too large package.json: %s bytes", tarEntry.getSize());
```
mcculls: prefer static constant over literal number

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+    distTags.put("latest", lastVersion);
+
+    final PackageRoot packageRoot = new PackageRoot(context.getRepository().getId(), rootJson);
+    hostedMetadataService.consumePackageRoot(packageRoot);
+    processedPackages = processedPackages + 1;
+    processedPackageVersions = processedPackageVersions + packages.size();
+  }
+
+  private Map<String, Object> extractPackageJson(final StorageFileItem file) throws IOException {
+    TarArchiveInputStream tarArchiveInputStream = new TarArchiveInputStream(file.getInputStream());
```
mcculls: stream doesn't appear to be closed

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/service/internal/orient/OrientMetadataStore.java
```diff
@@ -113,6 +128,46 @@ public void doStop() throws Exception {
...
+   * Performs DB backup, during which database is "frozen" for data modification attempts.
+   *
+   * @since 2.11
+   */
+  public void backupDatabase() throws IOException {
+    try (final ODatabaseDocumentTx db = db()) {
+      final File backupFile = createBackupFile();
+      log.info("Started npm DB backup to {} (all DB write operations are frozen)",
+          backupFile.getAbsolutePath());
+      final OutputStream outputStream = new FileOutputStream(backupFile);
```
mcculls: is `db.backup` responsible for closing the output stream?

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
cstamas: Ack, rename of these would be good, as they are badly named.

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
cstamas: From my empirical observation, that's right. But now am thinking about it, removal is not needed, as the dist-tags map is used only for "tag lookup".

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
cstamas: Yes, they are packageRoots and packageVersions. Rename imminent. 

This is the number of processed package roots.

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
cstamas: Ack, fixing it.

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/service/internal/orient/OrientMetadataStore.java
cstamas: Yes, db.backup wraps the stream (for buffering reasons), and closes it on return.

cstamas:
Back to WIP as tarballs have strange internal directories and current code is not enough to locate package.json within.

cstamas:
Sorted out tarballs structure, added IT.

mcculls:
+1

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/NpmHostedRepository.java
```diff
@@ -25,4 +25,11 @@
 {
   @Override
   HostedMetadataService getMetadataService();
+
+  /**
+   * Recreates npm metadata from deployed tarballs. During this operation, repository should be read only.
```
mrprescott: What does the return value mean?

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
```diff
@@ -0,0 +1,251 @@
...
+        }
+        final Version version = Version.parse(String.valueOf(versionJson.get("version")));
+        final Map<String, Object> dist = Maps.newHashMap();
+        dist.put("tarball", "generated-on-request"); // We lost original value, but does not matter as is NX hosted
+        dist.put("shasum", file.getRepositoryItemAttributes().get(StorageFileItem.DIGEST_SHA1_KEY));
+        versionJson.put("dist", dist);
+        versions.put(version, versionJson);
+      }
+      catch (Exception e) {
+        log.info("Failed to extract or malformed package.json from {}", item.getRepositoryItemUid());
```
mrprescott: Log the exception too?

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
```diff
@@ -0,0 +1,251 @@
...
+      log.debug("Examining TAR file");
+      final TarArchiveInputStream tarArchiveInputStream = new TarArchiveInputStream(gzipInputStream);
+      TarArchiveEntry tarEntry;
+      // npm uses tar option "--strip-components=1", so we need to strip first path element to have proper names
+      String tarEntryName;
+      do {
+        tarEntry = tarArchiveInputStream.getNextTarEntry();
+        tarEntryName = null;
+        if (tarEntry != null) {
+          // tar entry names are "pathelem1/pathelem2/.../filename"
```
mrprescott: Some of these methods are quite long; this seems like a candidate for a clearly named method that just parses out tar entry names (as the algorithm could then be unit-tested on its own, etc.)

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/internal/Version.java
```diff
@@ -0,0 +1,374 @@
...
+    /**
+     *
+     * Creates a Version from a string representation. Must match Version#FORMAT.
+     *
+     * @param version
+     * @return
+     */
+    public static Version parse(@Nonnull final String version) {
+        final Matcher matcher = Version.PATTERN.matcher(version);
+        if (!matcher.matches()) {
```
mrprescott: I keep getting told to use checkArgument(!matcher.matches(),"<%s> does not match format %s",version,Version.FORMAT) for this sort of thing. :P

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/internal/Version.java
```diff
@@ -0,0 +1,374 @@
...
+        final String special = matcher.group(5);
+        return new Version(major, minor, patch, separator, "".equals(special) ? null : special);
+    }
+
+    /**
+     * @param type
+     * @return next {@link Version} regarding specified {@link Version.Element}
+     */
+    public Version next(@Nonnull final Version.Element element) {
+        if (element == null) {
```
mrprescott: Same here, checkArgument(element==null,"null element")

mrprescott:
+1

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
```diff
@@ -0,0 +1,251 @@
...
+    final TreeMap<Version, Map<String, Object>> versions = Maps.newTreeMap();
+
+    // collect all the possible tarballs
+    final Collection<StorageItem> items = coll.list();
+    for (StorageItem item : items) {
+      if (!(item instanceof StorageFileItem)) {
+        // not a file, just skip it
+        continue;
+      }
+      if (!item.getName().endsWith(".tgz")) {
```
kellyrob99: Is this consistent for all tarballs? Not overly familiar with npm, but ".tar.gz" is equivalent

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
```diff
@@ -0,0 +1,251 @@
...
+        // not a tarball, just skip it
+        continue;
+      }
+      try {
+        final StorageFileItem file = (StorageFileItem) item;
+        final Map<String, Object> versionJson = extractPackageJson(file);
+        if (versionJson == null) {
+          // not found in tarball, just skip it but this time log it
+          log.info("Not found package.json in {}", file.getRepositoryItemUid());
```
kellyrob99: this file is mandatory for npm, isn't it?
https://docs.npmjs.com/misc/developers

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
```diff
@@ -0,0 +1,251 @@
...
+  {
+    liftAttributes(rootJson, packageJson);
+    final String version = String.valueOf(packageJson.get("version"));
+    final Map<String, Object> distTags = (Map<String, Object>) rootJson.get("dist-tags");
+    final Map<String, Object> versions = (Map<String, Object>) rootJson.get("versions");
+    distTags.put(version, version);
+    versions.put(version, packageJson);
+  }
+
+  private void liftAttributes(final Map<String, Object> higher, final Map<String, Object> lower) {
```
kellyrob99: Might be just me, but "lift" isn't terribly clear. Perhaps copyAttributes(from, to) would be closer?

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/internal/Version.java
```diff
@@ -0,0 +1,374 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+import org.apache.commons.lang.StringUtils;
+
+/**
+ * Version following semantic defined by <a href="http://semver.org/">Semantic Versioning</a> document.
+ *
+ * Copied from https://github.com/jeluard/semantic-versioning/blob/master/api/src/main/java/org/semver/Version.java
```
kellyrob99: Why are we copying this when the api jar is available in central? http://search.maven.org/#artifactdetails|org.semver|api|0.9.33|jar

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/internal/Version.java
kellyrob99: @mrprescott this is C&P from https://github.com/jeluard/semantic-versioning/blob/bca36dc2202f806ecc022b8b34f110eba830fe9e/api/src/main/java/org/semver/Version.java

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/service/internal/orient/OrientMetadataStore.java
```diff
@@ -113,6 +128,46 @@ public void doStop() throws Exception {
   }
 
   /**
+   * Performs DB backup, during which database is "frozen" for data modification attempts.
```
kellyrob99: Is the "freezing" automatic as part of the backup process? And what happens to other traffic trying to write data during this window?

mcculls: During backup all further write operations are locked, see http://www.orientechnologies.com/docs/2.0/orientdb.wiki/Backup-and-Restore.html#when-to-use-backup-and-when-export for more details

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/task/NpmDbBackupTask.java
```diff
@@ -0,0 +1,72 @@
...
+  public NpmDbBackupTask(final NexusConfiguration nexusConfiguration, final OrientMetadataStore orientMetadataStore) {
+    super(nexusConfiguration);
+    this.orientMetadataStore = checkNotNull(orientMetadataStore);
+  }
+
+  @Override
+  protected void doExecute(final List<NpmRepository> repositories) throws Exception {
+    final List<RepositoryWritePolicy> writePolicies = setWritePolicyReadOnly(repositories);
+    try {
+      orientMetadataStore.backupDatabase();
```
kellyrob99: what will happen if someone tries to add/delete repos while this is backing up?

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/task/NpmDbBackupTask.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Inject
+  public NpmDbBackupTask(final NexusConfiguration nexusConfiguration, final OrientMetadataStore orientMetadataStore) {
+    super(nexusConfiguration);
+    this.orientMetadataStore = checkNotNull(orientMetadataStore);
+  }
+
+  @Override
+  protected void doExecute(final List<NpmRepository> repositories) throws Exception {
+    final List<RepositoryWritePolicy> writePolicies = setWritePolicyReadOnly(repositories);
```
kellyrob99: parent NpmTaskSupport.doRun() method already does readOnly set/reset so I'm assuming it's not necessary to repeat it here?

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/task/NpmHostedMetadataRebuildTaskDescriptor.java
```diff
@@ -0,0 +1,59 @@
...
+ * @since 2.11
+ */
+@Named(NpmHostedMetadataRebuildTaskDescriptor.ID)
+@Singleton
+public class NpmHostedMetadataRebuildTaskDescriptor
+    extends AbstractScheduledTaskDescriptor
+{
+  public static final String ID = "NpmHostedMetadataRebuildTask";
+
+  static final String FLD_REPOSITORY_ID = "repositoryId";
```
kellyrob99: Let's hope we don't ever want to change the name of this property - defined/used in 106 different places in prod code. Isn't there someplace high enough up the dependency chain that we can define these commonly used keys for re-use?

testsuite/modern-testsuite/src/test/java/org/sonatype/nexus/testsuite/npm/rebuild/RebuildNpmMetadataIT.java
```diff
@@ -0,0 +1,79 @@
...
+  public RebuildNpmMetadataIT(String nexusBundleCoordinates) {
+    super(nexusBundleCoordinates);
+  }
+
+  @Test
+  public void rebuildNpmMetadata() throws Exception {
+    final NpmHostedRepository hostedRepository = createNpmHostedRepository(testMethodName());
+    final File hostRepositoryStorage = new File(nexus().getWorkDirectory(), "storage/" + hostedRepository.id());
+    FileUtils.copyFile(testData().resolveFile("commonjs-0.0.1.tgz"),
+        new File(hostRepositoryStorage, "commonjs/-/commonjs-0.0.1.tgz"), false);
```
kellyrob99: Is the '-' in the path significant?

testsuite/modern-testsuite/src/test/java/org/sonatype/nexus/testsuite/npm/rebuild/RebuildNpmMetadataIT.java
```diff
@@ -0,0 +1,79 @@
...
+    FileUtils.copyFile(testData().resolveFile("uppercase-0.1.1.tgz"),
+        new File(hostRepositoryStorage, "uppercase/-/uppercase-0.1.1.tgz"), false);
+
+    final Scheduler scheduler = client().getSubsystem(Scheduler.class);
+    scheduler.run("NpmHostedMetadataRebuildTask", null);
+    scheduler.waitForAllTasksToStop();
+
+    final File localDirectory = util.createTempDir();
+    // download package root of commonjs
+    {
```
kellyrob99: why is this code in a block please?

kellyrob99:
+1 with some comments/suggestions/questions but I'd definitely like to know why we would prefer Cut&Paste of semver code over consuming the same code through the published API

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/NpmHostedRepository.java
cstamas: Basically I followed the pattern we have for method like expireCaches, etc. `true` means there were packages processed (db changed). `false` means no packages processed (db unchanged), basically means repository is empty, at least contains no tarballs. Am totally neutral is return value needed at all or not.

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
cstamas: Ack, fixing

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
cstamas: From empirical experience (and some code reading), yes.
https://github.com/npm/npm-registry-client/blob/master/lib/publish.js#L78

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
cstamas: Yes, it is. The point of this (and why it is reported), to be foolproof it you happen to have a tarball in your npm repository does not have, or have a malformed npm package. Method `extractPackageJson` will return `null` in some cases, like `package.json` is:
* not present in tarball
* is present but is not a file
* is present as file, but lacks some mandatory fields.

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
cstamas: input parameter `packageJson` is being added to `rootJson`. So, it is "lift" one level upper (from child packageJson to parent rootJson). But I agree, am known to be terrible at naming things :)

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
cstamas: As after thought, the method `extractPackageJson` would need to be modified, to either return non-null map, or throw. fixing it,

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/task/NpmDbBackupTask.java
cstamas: Ack, fixing

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/task/NpmHostedMetadataRebuildTaskDescriptor.java
cstamas: This is fixed in NX3/Quartz tasks. 

In NX2 every task is free to declare field ID it likes, the properly is obeyed by UI (is dynamic), and the configurations are not interchanged.

testsuite/modern-testsuite/src/test/java/org/sonatype/nexus/testsuite/npm/rebuild/RebuildNpmMetadataIT.java
cstamas: Yes, that's part of npm repo spec, at least host NX hosted repository implements it.

cstamas:
@kellyrob99 Here is the POM of the library that contains the Version class:
https://repo1.maven.org/maven2/org/semver/api/0.9.33/api-0.9.33.pom

Despite the "API" name, the library is about diffing and byte code analysis of the inter-version binary compatibility, and it "just happens" to have implementation of SemVer in one single class. 

After a month, one would "WTF is npm plugin doing with java bytecode analysis"?

Personally, I think adding the library just for this one single class is overkill. Any other known good semver library maybe?

plugins/npm/nexus-npm-repository-plugin/src/main/java/com/bolyuba/nexus/plugin/npm/hosted/RecreateMetadataWalkerProcessor.java
kellyrob99: You are not alone ;)

cstamas:
Manually merged as d3de8a939a5cde3c0bccf4932ceb6e1899654682

PR #951, opened 2015-01-09
NEXUS-7883 - Role ordering now case sensitive

kellyrob99:
- update the ExtJS model with an appropriate sorter and enhance existing functional test to confirm behaviour

### DIFF
--- a/plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/model/Role.js
+++ b/plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/model/Role.js
@@ -23,7 +23,7 @@ Ext.define('NX.coreui.model.Role', {
     'id',
     'version',
     'source',
-    'name',
+    {name: 'name', sortType: 'asUCText'},
     'description',
     'readOnly',
     'privileges',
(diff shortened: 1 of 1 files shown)


plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/model/Role.js
```diff
@@ -23,7 +23,7 @@ Ext.define('NX.coreui.model.Role', {
     'id',
     'version',
     'source',
-    'name',
+    {name: 'name', sortType: 'asUCText'},
```
jdillon: minor style note, may be nicer to uniformlly use object config and not mix, /me shrugs

jdillon:
+1 with minor comment

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/model/Role.js
kellyrob99: 01dd9de9cd00ee427f9a7850cbd46d9c6a599ca7

djsauble:
+1

PR #952, opened 2015-01-09
Creating new items is now a drilldown experience

djsauble:
Before:
http://take.ms/meF2G

After:
http://take.ms/JzmlR

Related PRs:
https://github.com/sonatype/nexus-pro/pull/468
https://github.com/sonatype/nexus-bundles/pull/121


adreghiciu:
I spent quite some hours today looking into capabilities problem. I found out that staging profiles have the same problem.
I could not find a solution but it is related to the sliding/animation. If you replace the slidePanel with https://gist.github.com/adreghiciu/0914c4829e019697303f everything will work as expected.

Actually I think we should get rid of sliding animation as it only adds complexity without any real benefit beside the nice animation.

djsauble:
@adreghiciu I just figured this out. When you resize the panel, the problem resolves itself (http://take.ms/nt9Id).

This made me realize that the problem was with syncSizeToOwner(). This function is called once to do initial layout of the drilldown, and then an additional call whenever the window is resized.

The problem is that the initial call was made _before_ rendering of children was complete. To fix this, I [put the initial call to syncSizeToOwner() inside an afterrender event handler](https://github.com/sonatype/nexus-oss/commit/b0c7311629a93fdc2fff6de01153d54d61164987). Problem solved.

The benefit of the animation is that it enforces the fact that the user is in a workflow, and can move back and forth at will. Essentially, a “soft” navigation, where the state of the parent pages is preserved (i.e. selected row, scroll position, filter, etc…).

If we remove the animation, people will expect a “hard” navigation when they go back (essentially a page refresh). In that sense, the animation helps set a more realistic expectation of what will happen during navigation.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/drilldown/Drilldown.js
```diff
@@ -197,5 +207,264 @@ Ext.define('NX.view.drilldown.Drilldown', {
...
+        afterrender: me.syncSizeToOwner
+      });
+    }
+
+    me.callParent(arguments);
+  },
+
+  /**
+   * Given N drilldown items, this panel should have a width of N times the current screen width
+   */
```
jdillon: please indicate intended visibility @public @private etc

djsauble: Done.

adreghiciu:
Did a local test and looks much better now. Good work.
One small problem which can be solved later on probably. With the add dialogs the first field always had focus. With drill down that does not happen anymore.

adreghiciu:
Another thing: I find confusing the change related to placement of cancel/create buttons, which does not match the save/cancel in update or settings panels

jdillon:
I agree with @adreghiciu that the button placement is odd, any reason why we invert the order for add @djsauble ?

jdillon:
Noticed a few problems with ssl, capabilities

djsauble:
@jdillon SSL certificate creation is now fixed. I wasn’t able to reproduce any problems when creating a new capability.

djsauble:
@jdillon @adreghiciu I’d like us to adopt the way OSX orders buttons. Buttons are right-aligned. Primary action on the right, secondary actions to the left. This gives the action more prominence (instead of being squashed against the busy primary nav, like it is today).

That said, I didn’t do a thorough job of switching to this order (and the buttons weren’t right-aligned). I’ve reverted that change in this PR and will tackle it at a later point.

I think we should merge what we have. I’ve noted the focus problem in [NEXUS-7935](https://issues.sonatype.org/browse/NEXUS-7935).

Reference:
https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/WindowDialogs.html#//apple_ref/doc/uid/20000957-CH43-SW1

jdillon:
@djsauble why?  I thought we had a chat about this yesterday about why the buttons are all left-aligned and the problems that right-aligned puts into the overall design.

djsauble:
@jdillon If we make buttons right-aligned and leave the rest of the design, as is, yeah–it’s a stupid idea. This involves larger changes to the design, particularly forms (see below), which is why I think we punt this for later.

ldap - sonatype nexus 2015-01-14 11-21-13

adreghiciu:
Go on and merge

djsauble:
Manual merge.

PR #955, opened 2015-01-14
[cma] Elasticsearch

adreghiciu:


components/nexus-elasticsearch/src/main/java/org/sonatype/nexus/elasticsearch/internal/ElasticSearchClientProvider.java
```diff
@@ -0,0 +1,84 @@
...
+    extends StateGuardLifecycleSupport
+    implements Provider<Client>
+{
+
+  private final File appDir;
+
+  private Node node;
+
+  @Inject
+  public ElasticSearchClientProvider(final @Named("${nexus-app}") File appDir) {
```
jdillon: Should use ApplicationDirectories

components/nexus-elasticsearch/src/main/java/org/sonatype/nexus/elasticsearch/internal/ElasticSearchClientProvider.java
```diff
@@ -0,0 +1,84 @@
...
+  {
+
+    @Inject
+    ElasticSearchClientProvider provider;
+
+    @PostConstruct
+    void start() throws Exception { provider.start(); }
+
+    @PreDestroy
+    void stop() throws Exception { provider.stop(); }
```
jdillon: Lifecycle probably isn't tied to the guice container, probably needs to be managed by app in which case these workarounds can/shold go away.

jdillon: I'd probably move this entierlly to nexus-core in org.sonatype.nexus.internal.elasticsearch, wire up to NX events and use the ApplicationDirectories helper.  This is really about hooking up the ES bits to nexus anyways so thats probably better than what is being done here.

jdillon:
merged into nexus-cma2

components/nexus-elasticsearch/src/main/java/org/sonatype/nexus/elasticsearch/internal/ElasticSearchClientProvider.java
adreghiciu: I did not used that as I did not want to depend on nexus-core. Yet by moving to nexus-core this will be possible to do

PR #960, opened 2015-01-16
Repositories CMA UI

adreghiciu:


jdillon:
Generally think this is good, but having issues with the naming of things.  I think we'll likely want to refine this UI as we go an not rewrite it and I think putting CMA all over the namespace is pretty icky.

Though atm, I'm struggling for a better means to stand up parallel UI bits for "repository" things which are for new (CMA) and not conflict with old (legacy) bits.

Also this should probably merge into nexus-cma2 not master.

adreghiciu:
Pull was against master as cma2 branch did not contain the latest changes made by Daniel about "new" drill down.

adreghiciu:
Replaced by sonatype/nexus-oss/pull/974

PR #962, opened 2015-01-16
Proper render About in Capabilities New

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-7854

CapabilityStatus was not really necessary to be changed but the panel is same as about


djsauble:
+1

PR #965, opened 2015-01-16
NEXUS-7901 groovy upgrade

kellyrob99:
Upgrade groovy to newer version as only groovy 2.8+ is actually workable in Java8, although earlier versions may/may not work depending on what you're tyring to do (among other reasons to upgrade).
Challenges:
- resolution of visible fields/methods has changed, leading to problems where Guice intercepted methods with Closures calling private methods can no longer find them

Options:
- @CompileStatic code does not appear to suffer from the lookup problem, but forces us to abandon some of the Groovier syntax and dynamic typing features. Also, it too seems to have some strange buggy behaviour, like the GDK method lookup failure explicitly mentioned later
- remote private modifier from affected methods and apply @PackageScope to keep the visibility as tight as possible

@PackageScope option gets my :+1: , looking forward to votes from the rest of the team on a path forward.

### DIFF
--- a/buildsupport/groovy/pom.xml
+++ b/buildsupport/groovy/pom.xml
@@ -16,9 +16,6 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 
   <modelVersion>4.0.0</modelVersion>
-  <properties>
-    <groovy.version>2.3.7</groovy.version>
-  </properties>
 
   <parent>
     <groupId>org.sonatype.nexus.buildsupport</groupId>
@@ -30,6 +27,10 @@
   <name>${project.groupId}:${project.artifactId}</name>
   <packaging>pom</packaging>
 
+  <properties>
+    <groovy.version>2.3.7</groovy.version>
+  </properties>
+  
   <dependencyManagement>
     <dependencies>
 
--- a/buildsupport/groovy/pom.xml
+++ b/buildsupport/groovy/pom.xml
@@ -16,6 +16,9 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 
   <modelVersion>4.0.0</modelVersion>
+  <properties>
+    <groovy.version>2.3.7</groovy.version>
+  </properties>
 
   <parent>
     <groupId>org.sonatype.nexus.buildsupport</groupId>
@@ -33,13 +36,13 @@
       <dependency>
         <groupId>org.codehaus.groovy</groupId>
         <artifactId>groovy-all</artifactId>
-        <version>2.1.5</version>
+        <version>${groovy.version}</version>
       </dependency>
 
       <dependency>
         <groupId>org.codehaus.groovy</groupId>
         <artifactId>groovy</artifactId>
-        <version>2.1.5</version>
+        <version>${groovy.version}</version>
       </dependency>
 
       <!--
--- a/plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/MavenUploadComponent.groovy
+++ b/plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/MavenUploadComponent.groovy
@@ -105,6 +105,7 @@ extends DirectComponentSupport
     }
   }
 
+  @PackageScope
   static ArtifactStoreRequest createRequest(final FileItem file,
                                                     final MavenRepository mavenRepository,
                                                     final UploadContext uploadContext,
@@ -140,7 +141,7 @@ extends DirectComponentSupport
 
   @PackageScope
   static UploadContext createUploadContext(final Map<String, String> params,
-                                                   final Map<String, FileItem> files)
+      final Map<String, FileItem> files)
   {
     UploadContext context = new UploadContext(
         pomAvailable: false,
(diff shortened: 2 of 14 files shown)


buildsupport/groovy/pom.xml
```diff
@@ -16,6 +16,9 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 
   <modelVersion>4.0.0</modelVersion>
+  <properties>
+    <groovy.version>2.3.7</groovy.version>
```
kellyrob99: should be able to got to 2.3.9, but they do recommend using matching versions of eclipse-batch

plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/PrivilegeComponent.groovy
```diff
@@ -104,20 +106,22 @@ extends DirectComponentSupport
...
-              type: TargetPrivilegeDescriptor.TYPE,
-              readOnly: false,
-              properties: [
-                  (ApplicationPrivilegeMethodPropertyDescriptor.ID): method,
-                  (TargetPrivilegeRepositoryTargetPropertyDescriptor.ID): privilegeXO.repositoryTargetId,
-                  (TargetPrivilegeRepositoryPropertyDescriptor.ID): repositoryId,
-                  (TargetPrivilegeGroupPropertyDescriptor.ID): groupId
-              ]
+    ['create', 'read', 'update', 'delete'].each { String method ->
+      created << asPrivilegeXO(
```
kellyrob99: Ran into a real typing issue here where we filled the List<PrivilegeXO> with Privilege objects and had it work mostly due to the fact that they are essentially copies of each other. Should be fixed regardless of decisions made regarding this PR.

plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/PrivilegeComponent.groovy
```diff
@@ -162,7 +166,7 @@ extends DirectComponentSupport
       else if (key == 'repositoryId' || key == 'repositoryGroupId') {
         if (!privilege.repositoryId) { // check not already mapped
           privilege.repositoryId = value
-          if (value && value != '*' && !value.allWhitespace) {
+          if (value && value != '*' && !value.isAllWhitespace()) {
```
kellyrob99: Compile failure using property access instead of method

plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/PrivilegeComponent.groovy
```diff
@@ -52,6 +53,7 @@ import javax.validation.groups.Default
 @Named
 @Singleton
 @DirectAction(action = 'coreui_Privilege')
+@CompileStatic
```
kellyrob99: example of @CompileStatic on this class, chosen because it required far less change than others; RepositoryComponent for instance reported a boatload of compile errors with this applied

plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/RoleComponent.groovy
```diff
@@ -170,7 +171,8 @@ extends DirectComponentSupport
     securitySystem.getAuthorizationManager(DEFAULT_SOURCE).deleteRole(id)
   }
 
-  private static RoleXO asRoleXO(Role input) {
+  @PackageScope
```
kellyrob99: applying this and removing private modifier works, and is the lowest barrier to fixing the problem while maintaining some privacy

buildsupport/groovy/pom.xml
mcculls: Minor: typically we put the properties section just above the dependencyManagement section.

plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/MavenUploadComponent.groovy
```diff
@@ -104,7 +105,7 @@ extends DirectComponentSupport
     }
   }
 
-  private static ArtifactStoreRequest createRequest(final FileItem file,
+  static ArtifactStoreRequest createRequest(final FileItem file,
```
mcculls: Any particular reason this one method doesn't have `@PackageScope`?

buildsupport/groovy/pom.xml
jdillon: yar, should be moved down to normalize with all other poms in nx projects.

plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/MavenUploadComponent.groovy
```diff
@@ -137,7 +138,8 @@ extends DirectComponentSupport
     return result
   }
 
-  private static UploadContext createUploadContext(final Map<String, String> params,
+  @PackageScope
+  static UploadContext createUploadContext(final Map<String, String> params,
```
mcculls: File needs reformatting, as multi-line parameters are no longer aligned

mcculls:
+1

buildsupport/groovy/pom.xml
kellyrob99: No problemo 371fc0a2c0a22114b031dfb3bd33791341fee218

On Mon, Jan 19, 2015 at 11:07 AM, Stuart McCulloch <notifications@github.com

plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/MavenUploadComponent.groovy
kellyrob99: Nope, just missed it. 57b11c0d07cbc4fa98213f0e066ec59f10ea7e93

jdillon:
+1

PR #969, opened 2015-01-18
Expose real permission

jdillon:
Exposes the "real" shiro permission in the UI.

screen shot

Detail view also now shows the priv.properties asis:

screen shot2

Will refine the UI around this later I'm sure.


jdillon:
There is duplicate information (method) which applies only to a sub-set of privileges, as well as more data in the grid that only applies to others, which shows a bit of anti-pattern for grid data display.

Had to use "Real Permission" and "realPermission" as the underlying model already has a "permission" attribute, but this only applies to the "Application" aka "method" privileges.

Some notes in this PR as well to clean up this so we can properly scale the UI and data-model for various privilege types.

Also a minor note, is we have an inconsistent permission naming scheme, where most use "nexus:" there are more that use other prefixes, like:

* apikey:
* security:

This does add additional confusion and complication when using wildcard permissions and should probably be cleaned up.  The "security:" bits are likely side-effect of all that code existing _outside_ of nexus, but only ever being used by nexus and now that these are merged, the permissions should probably have been adjusted to use "nexus:" prefix.

jdillon:
This further shows the anti-pattern if mixed types in the grid with columns of blank fields:

https://www.evernote.com/shard/s18/sh/00d06fec-8f5e-4041-9e5a-6b1d775fd3e0/371ee4fd69cf9fa5c30a94f37570203b

jdillon:
Another issue I think we have had for a long time is id+name, which are about the same thing, need to be kept in sync, and force us to pass along both strings to the UI, but in code the ids would be what would be logged w/o having more complexity to go back to find the "name".

https://www.evernote.com/shard/s18/sh/0abed22c-a7eb-4645-b4a4-e617b0e8a4fe/8adcff52c0d0b94c13744420921ee362

This may be much simpler to drop the "name" and instead let the "description" be the _human readable/comprehendible_ text instead.

IMO the "id" values are just as descriptive as the "name" values in nearly all cases.  And folks trying to understand the use of permissions, in code and in logs may find this much more useful than some rather b/s "name" that only shows up in the UI.

jdillon:
FTR this IMO is much better use of the UI to display the relevant information:

https://www.evernote.com/shard/s18/sh/af4e3a31-c419-4ab3-b56d-aa56c3509140/9284c338dd6b18c4e8ed603f44663ff0

... and not waste space.

djsauble:
+ “Type” column is too wide.
+ Why “Real Permission” instead of “Shiro Permission”, “Shiro Permission String” or similar?
+ No default sort column? Should be ID (ascending).
+ I get the desire to put the icon next to the “Type” column, but it’s not the most important column in this table. Order should be ID -> Type -> Description -> Real Permission.
+ Detail view is missing the “Real Permission” and “Type” attributes.
+ “Target” is called “Repository target” in the detail view. We should pick one name and stick with it. Personally, I vote for “Repository target”.

djsauble:
This is what I’d expect the table to look like by default (Ignore the blank “Permission String” column. I needed to rebuild or something.)

screen shot 2015-01-19 at 9 24 30 am

djsauble:
+1

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/controller/Privileges.js
```diff
@@ -140,23 +140,24 @@ Ext.define('NX.coreui.controller.Privileges', {
...
+  //
+  //  targetCombo.setValue(undefined);
+  //  if (repositoryId) {
+  //    format = repositoryIdCombo.getStore().getById(repositoryId).get('format');
+  //    targetCombo.getStore().filterByFormat(format);
+  //  }
+  //  else {
+  //    targetCombo.getStore().removeFilterByFormat();
+  //  }
+  //},
```
jdillon: pending some info form @adreghiciu about this

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/controller/Privileges.js
```diff
@@ -92,14 +92,18 @@ Ext.define('NX.coreui.controller.Privileges', {
...
         'nx-coreui-privilege-add-repositorytarget form': {
           submitted: me.onSettingsSubmitted
-        },
-        'nx-coreui-privilege-add-repositorytarget #repositoryId': {
-          change: me.filterRepositoryTargets
         }
+        // FIXME: What does this do?
+        //'nx-coreui-privilege-add-repositorytarget #repositoryId': {
+        //  change: me.filterRepositoryTargets
+        //}
```
jdillon: Pending some info from @adreghiciu  about this

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/controller/Privileges.js
```diff
@@ -208,13 +209,14 @@ Ext.define('NX.coreui.controller.Privileges', {
...
     var me = this,
-        description = me.getDescription(model);
+        name = model.get('name');
 
     NX.direct.coreui_Privilege.remove(model.getId(), function (response) {
       me.loadStore();
       if (Ext.isObject(response) && response.success) {
         NX.Messages.add({
-          text: 'Privilege deleted: ' + description, type: 'success'
+          text: NX.I18n.format('ADMIN_PRIVILEGES_MESSAGE_DELETED', name),
```
kellyrob99: minor but given it's single use we could ditch the temp variable and just call it here?

jdillon:
http://bamboo.s/browse/NX-OSSF500-1 :white_check_mark:

adreghiciu:
+1 after the commented functionality is put back

cstamas:
+1

jdillon:
Will need to update https://github.com/sonatype/nexus-bundles/blob/master/testsuite/functional-testsuite/src/test/ft-resources/testsuite/security/110_security_privilege.t.js

adreghiciu:
+1

jdillon:
manually merged

PR #976, opened 2015-01-19
Fixes drilldown bugs

djsauble:
Related PR: https://github.com/sonatype/nexus-pro/pull/472

Related issues:

https://issues.sonatype.org/browse/NEXUS-7660
https://issues.sonatype.org/browse/NEXUS-7759
https://issues.sonatype.org/browse/NEXUS-7951
https://issues.sonatype.org/browse/NEXUS-7948

Changes:
+ Reset the bookmark when a model is deleted
+ Clear stale data in create screens
+ Fix error on cancel when adding a new SSL trusted key
+ Restore tab navigation (clicking the “back” button goes to the previous tab)
+ Detail tabs for task don’t disappear on refresh
+ “About” tab for capability is populated on refresh
+ Restores broken “Change password” functionality


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Drilldown.js
```diff
@@ -40,9 +40,16 @@ Ext.define('NX.controller.Drilldown', {
 
   permission: undefined,
 
+  /**
+   * @override
```
jdillon: override here seems odd, I'd expect to see sub-classes use this anno, but not here on the def.  Here public/protected probably more useful.  Unsless I'm missing that the super-class has a getDescription which this class is overriding.

jdillon:
+1, minor details about use of jsduck annos

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Drilldown.js
djsauble: No, you’re right. @protected is what I wanted here.

I’ll add @override to the classes which inherit this.

djsauble:
Manual merge.

PR #987, opened 2015-01-22
Do not select a logger after it is created

djsauble:
None of the other drilldown views behave this way, so this change makes the behavior of logger consistent with the rest of the web UI.

Issue: https://issues.sonatype.org/browse/NEXUS-7934

Before: http://take.ms/speiI

After: http://take.ms/NS1z3


cstamas:
+1

adreghiciu:
+1

djsauble:
Manual merge https://github.com/sonatype/nexus-oss/commit/c7fbfd04717ecf3422a8ab09521a4a0218aab908

PR #988, opened 2015-01-22
NEXUS-7962 Forces the tab panel body to sit underneath the tab bar

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-7962

Before:
users - sonatype nexus 2015-01-21 17-48-07

After:
users - sonatype nexus 2015-01-21 17-47-30


cstamas:
+1

kellyrob99:
Seems a bit strange that we have to override extjs styles to get 'correct' behaviour - is this a bug in their css or just a collision with our usage?
If it's their bug, we should raise an issue and propose our fix, otherwise +1

jdillon:
Did we undo the previous tab css changes?  I had to do some hackery to make those work (and it sill had issues)

djsauble:
Not intentionally. I had to hack your hacks to get pills to show up properly, but everything should display properly now.

jdillon:
@djsauble okay, @kellyrob99 this may be simplified post upgrade to extjs5 which uses pills, but may need a bit of magichackery until then.

+1

djsauble:
Manual merge https://github.com/sonatype/nexus-oss/commit/6978312bea9b7fb306a3f90d84d235a288800932

PR #991, opened 2015-01-22
NEXUS-7939 - Privilege add validates with incorrect message?

kellyrob99:
- allow empty values for repositoryId, which gets translated to a '*' permission element for repos

https://issues.sonatype.org/browse/NEXUS-7939

### DIFF
--- a/plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/ExtDirect.js
+++ b/plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/ExtDirect.js
@@ -75,7 +75,11 @@ Ext.define('NX.controller.ExtDirect', {
     if (message) {
       NX.Messages.add({text: message, type: 'warning'});
     }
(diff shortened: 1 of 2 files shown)


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/ExtDirect.js
```diff
@@ -75,7 +75,11 @@ Ext.define('NX.controller.ExtDirect', {
     if (message) {
       NX.Messages.add({text: message, type: 'warning'});
     }
-    me.logDebug(transaction.action + ':' + transaction.method + " -> " + (message ? 'Failed: ' + message : 'OK'));
+    var logMsg = transaction.action + ':' + transaction.method + " -> " + (message ? 'Failed: ' + message : 'OK');
```
kellyrob99: adding this as validation errors can be included even when 'OK' is logged to the console

adreghiciu:
+1

jdillon:
+1

PR #992, opened 2015-01-23
NEXUS-7935 When you click a drilldown, focus on the new panel

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-7935

**The Good**: Focus works properly… except for grids… they don’t seem to take focus, so you can’t actually navigate a wizard with the keyboard alone. :-(

**The Bad**: Drilldown animation no longer works. This makes me sad, but I wasn’t able to fix [this issue](http://take.ms/UFZkR) without reverting to a standard card layout (which, unfortunately, doesn’t have support for animation).

Before: http://take.ms/keboT

After: http://take.ms/5tQLK


djsauble:
Superseded by https://github.com/sonatype/nexus-oss/pull/998, which fixes the same issues while preserving animation.

PR #993, opened 2015-01-23
NEXUS-7995: Bump HttpClient et al. (2.11.x)

cstamas:
Update HttpClient and HttpCore to:
* HttpClient 4.3.6, [release notes](https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12310360&version=12327489)
* HttpCore 4.3.3, [release notes](https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12310340&version=12326260)

Issue
https://issues.sonatype.org/browse/NEXUS-7995

CI
http://bamboo.s/browse/NX-OSSF526

### DIFF
--- a/buildsupport/httpclient/pom.xml
+++ b/buildsupport/httpclient/pom.xml
@@ -33,7 +33,7 @@
       <dependency>
         <groupId>org.apache.httpcomponents</groupId>
         <artifactId>httpclient</artifactId>
-        <version>4.3.5</version>
+        <version>4.3.6</version>
         <exclusions>
           <exclusion>
             <groupId>commons-logging</groupId>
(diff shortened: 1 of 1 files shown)


buildsupport/httpclient/pom.xml
```diff
@@ -33,7 +33,7 @@
       <dependency>
         <groupId>org.apache.httpcomponents</groupId>
         <artifactId>httpclient</artifactId>
-        <version>4.3.5</version>
+        <version>4.3.6</version>
```
kellyrob99: extract a property for this?

kellyrob99:
+1

buildsupport/httpclient/pom.xml
jdillon: why?  this is used in 2 places? If it grows to more perhaps, but we don't need properties to manage versions for all dependencies, especially not in these tiny buildsupport modules.

cstamas:
http://bamboo.s/browse/NX-OSSF526-1 :+1:

PR #1002, opened 2015-01-28
[NEXUS-7647] Implement createrepo/mergerepo in java

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-7647

CI : http://bamboo.s/browse/NX-OSSF521-3 :+1:


jdillon:
+1 to move forward with saner yum impl

FTR wasn't sure why we needed to complicate this change with database bits, but ... it is what it is.

adreghiciu:
To generate the yum repository we need to crack open every rpm and extract the yum metadata.
This need the metadata on each rpm deploy as that is the info necessary to generate metadata.
In order to avoid opening the rpm files I decided to use a db to store the extracted metadata (cache it) so is necessary to extract the info only for newly added rpm.

Createrepo is doing something similar just that is caching that info on disk in a file. We could do the same but I found more useful to have the info it in a db as it should be faster to access it. The YumStore imply that keeps the yum metadata in db now can be reimplemented in other ways if we consider that using the db is not what we want. It could even re-extract metadata from rpm every time.
When we do the CMA imply we could decide to store the extracted metadata in component metadata as json.

plugins/yum/nexus-yum-repository-plugin/src/main/java/org/sonatype/nexus/yum/internal/YumHostedImpl.java
```diff
@@ -107,10 +110,10 @@ public YumHostedImpl(final TaskScheduler nexusScheduler,
 
     this.aliases = Maps.newHashMap();
 
-    this.baseDir = RepositoryUtils.getBaseDir(repository);
-
     this.yumGroupsDefinitionFile = null;
 
+    this.yumStore = checkNotNull(yumStoreFactory).create(repository.getId());
```
cstamas: Not much of value in checkNotNull here, as same effect with same NPE would be thrown if not used at all?

cstamas:
looks good

jdillon:
@adreghiciu understood, thanks for the detail.

adreghiciu:
Manually merged https://github.com/sonatype/nexus-oss/commit/caacdc5407c0ae59ba67822fa4782587f9d8dfb5

PR #1020, opened 2015-02-02
Restores drilldown carets missing due to ca8b83d06

djsauble:
When columns are added statically (i.e. as part of the grid definition), `add` is never triggered. The only way to add drilldown carets to these grids is through the `viewready` event, which I’ve reinstituted in this PR.

Before:
plugins - sonatype nexus 2015-02-02 10-36-01

After:
plugins - sonatype nexus 2015-02-02 10-35-24


adreghiciu:
+1

cstamas:
+1

djsauble:
Manual merge. https://github.com/sonatype/nexus-oss/commit/b42e60267829d1ec104dc8eda193ae6f51d4a5d7

PR #1024, opened 2015-02-05
[NEXUS-8011] Select capability type before showing add panel

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8011
http://screencast.com/t/5C8FPOKnG


plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/controller/Capabilities.js
```diff
@@ -279,9 +296,9 @@ Ext.define('NX.coreui.controller.Capabilities', {
 
     if (list) {
       me.getCapabilityTypeStore().load(
-        function() {
-          me.reselect();
-        }
+          function() {
```
kellyrob99: is 4 spaces the desired format? been using 2 spaces in js for so long it looks... strange

kellyrob99:
+1 and tnx @adreghiciu for including the screencast, it really helps

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/controller/Capabilities.js
adreghiciu: I'm also using 2 spaces. Dunno why IDEA reformat those lines

adreghiciu: Found it: http://screencast.com/t/2uWRgBkl
And that is in out standard code style https://github.com/sonatype/codestyle/blob/master/sonatype-idea.xml

kellyrob99: Tnx for the followup

djsauble:
Can we show the first line of the description as a second column in that grid? Scanning a list is much faster than expanding each item individually.

If the description exceeds the available space, can we use a tooltip to show the rest the way Repository Health Check does?

screen shot 2015-02-05 at 11 12 34 am

adreghiciu:
Agree that a tooltip will be better.

adreghiciu:
Updated to use tooltip and show first part of about text. Note that about text is an HTML so I show only the part till first dot (.)

djsauble:
+1. Looks fantastic.

jdillon:
+1... missing > indicator though?

djsauble:
That’s true. I can’t see why they’re missing though.

PR #1027, opened 2015-02-07
[NEXUS-8011] Select task type before showing add panel

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8012
Before: http://take.ms/CndsT
After: http://take.ms/HdliE


adreghiciu:
+1

kellyrob99:
+1

PR #1028, opened 2015-02-07
[NEXUS-8117] Fixes intermittent problem with drilldown

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8117


jdillon:
unable to reproduce with this change

PR #1030, opened 2015-02-09
Tweaks readonly treatment to fix a few edge cases

djsauble:
**Field containers have a different background color when read only**

Before:
roles - sonatype nexus 2015-02-09 10-18-18

After:
roles - sonatype nexus 2015-02-09 10-15-44

**Log viewer bleeds into the secondary navigation**

Before:
screen shot 2015-02-09 at 10 20 16 am

After:
log viewer - sonatype nexus 2015-02-09 10-16-12

**Combo boxes get read-only treatment when not read-only**

Before:
screen shot 2015-02-09 at 10 23 47 am

After:
screen shot 2015-02-09 at 10 24 33 am


jdillon:
+1

adreghiciu:
+1

PR #1031, opened 2015-02-10
NEXUS-7758 Consolidate tabs that are light on content

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-7758
Related to: https://github.com/sonatype/nexus-pro/pull/481

**LDAP**

Before: http://take.ms/FR9MY
After: http://take.ms/YhMhD

**Search**

Before: http://take.ms/o4e3s
After: http://take.ms/e9WzR

**Capabilities**

Before: http://take.ms/kWwdm
After: http://take.ms/0wdI3


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/ext/panel/SubsectionPanel.js
```diff
@@ -0,0 +1,62 @@
...
+
+      me.suspendEvents();
+      me.remove(component, false);
+
+      me.items.each(function (item) {
+        var thatTitle = item.title || '',
+          thatWeight = item.weight || 1000;
+
+        if (thisWeight < thatWeight
+          || (thisWeight === thatWeight && thisTitle < thatTitle)) {
```
kellyrob99: unsure why we compare thisTitle < thatTitle?

kellyrob99:
+1 I like the higher density of info
Something to keep in mind for this particular use case is that once CLM is enabled the content changes, and the layout is different. 
screen shot 2015-02-12 at 10 25 22 am
This already looks better than in 2.11, where the CIP(Component Info Panel) is pretty squashed by default(when it opens the text and graphics by default stack one on top of another until you manually expand the panel, and the lower portion is initially hidden).
screen shot 2015-02-12 at 10 32 37 am

jdillon:
Initial reaction is ... better, but likely will not scale very well.  I think instead at least 2 tabs may be better.  One for a summary (which has high-level format details, snips of examples like that mvn xml, and may be the clm chiclets) and then another tab for the clm policy report.

djsauble:
@jdillon What aspect of scaling are you concerned about?

The “Component info” title is always visible at 1024x768 (unless there’s some gnarly edge-case I’m missing), and scrolling down is at least as fast as clicking a separate tab (and mental orientation is faster, since part of the license analysis and security issue tables are exposed, making it obvious what “Component info” means).

djsauble:
@kellyrob99 What configuration on the CLM side do I need in order to see that alternate view? I connected to the CLM server, but the “Component info” section now shows as blank.

screen shot 2015-02-12 at 3 41 02 pm

jdillon:
@djsauble information scale, the info/maven bits could become much more details, and different formats have different details to show.  The CLM policy report information is also already pretty detailed and can get more detailed.  So I think it would be generally better to have 2 tabs to cope with this.

In otherwords with all of these on the same view, its very busy and could get even busier when we add download links and other features.  So putting everything into one view will not scale very well as we flesh this out with more details/options/features.

djsauble:
Thanks for clarifying, makes sense to me.

kellyrob99:
Should be functional as soon as
https://github.com/sonatype/nexus-pro/pull/484 is merged in

joedragons:
I like it.  Might consider underlining the headers to provide more emphasis on different areas.
I'd like to see a busier page as well, if such an example exists...maybe it doesn't and we're keeping busy areas multiple tabs.

Or maybe that's what Kelly and Jason are speaking towards.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/ext/panel/SubsectionPanel.js
djsauble: I think it’s so tabs with the same weight are sorted alphabetically by default.

@adreghiciu ?

djsauble:
@joedragons @jdillon I say we not borrow trouble from the future. Let’s wait until Search details become more busy/complicated, and reintroduce tabs at that point.

For what it’s worth, it’s trivial to mix and match tabs and subsections, as shown in the Capabilities feature.

jdillon:
/me shrugs

I'm on the fence about deferring, vs setting the expectation now.

joedragons:
NP, I defer to y'all.
I am honestly not sure the state of search.  If there's more work to be
done, your logic makes sense.

PR #1044, opened 2015-02-16
NEXUS-8076: Ignore IOEx when gathering volume infos

cstamas:
In some cases IOEx is thrown by FileStore objects, that are inspecting some properties of it.

Issue
https://issues.sonatype.org/browse/NEXUS-8076

CI
http://bamboo.s/browse/NX-OSSF579


kellyrob99:
+1

plugins/internal/nexus-atlas-plugin/src/main/java/org/sonatype/nexus/atlas/internal/SystemInformationGeneratorImpl.groovy
```diff
@@ -109,14 +109,25 @@ class SystemInformationGeneratorImpl
...
+        try {
+          return callMe.call()
+        }
+        catch (e) {
+          if (!e.class.isAssignableFrom(IOException.class)) {
+            throw e
+          }
+          return defaultValue
+        }
+      }
```
jdillon: simplify this to just eat all exceptions, no reason to special case IOException here.
probably log trace the exception as well.

cwilper:
+1

PR #1047, opened 2015-02-19
NEXUS-8070: Remove of forgotId and forgotPw

cstamas:
As it performs badly on external realms.

Issue
https://issues.sonatype.org/browse/NEXUS-8070


jdillon:
why not remove the code?

cstamas:
@jdillon I just went with the "simplest solution w/ least impact", as one of my mates described in issue comment :smile:

jdillon:
impact of leaving in commented code that is intended to be removed seems higher than removing it in terms of long term code management.  Only reason we'd leave it in is if we planned to bring it back, which we don't.

jdillon:
anyways, +1 if this removes the crap from the UI and doesn't break any of this very fragile JS

cstamas:
Code removed and locally tested. Extra eyeballs would be nice, to ensure did not break anything.

kellyrob99:
+1

PR #1050, opened 2015-02-20
[NEXUS-7887][CMA] Negative cache

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-7887


buildsupport/testing/pom.xml
```diff
@@ -136,6 +136,12 @@
         <artifactId>junit</artifactId>
         <version>4.11</version>
       </dependency>
+
+      <dependency>
+        <groupId>javax.transaction</groupId>
+        <artifactId>javax.transaction-api</artifactId>
+        <version>1.2</version>
+      </dependency>
```
jdillon: This isn't really a testing dependency, though it may only be used as a test, I'd move this to another buildsupport group.

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/negativecache/NegativeCacheFacetImpl.java
```diff
@@ -0,0 +1,153 @@
...
+      }
+      cache = null;
+    }
+  }
+
+  @Override
+  @Guarded(by = STARTED)
+  public Status get(final NegativeCacheKey key) {
+    checkNotNull(key);
+    if (cache != null) {
```
kellyrob99: Seems like an invalid state if the cache was null?

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/negativecache/NegativeCacheHandler.java
```diff
@@ -30,35 +36,33 @@
...
-      return HttpResponses.notFound();
+    Response response;
+    Status status = negativeCache.get(key);
+    if (status == null) {
+      response = context.proceed();
+      if (isNotFound(response)) {
+        negativeCache.put(key, response.getStatus());
+      }
+      else if (response.getStatus().isSuccessful()) {
+        negativeCache.invalidate(key);
```
kellyrob99: do we need to invalidate in this case? we could only get here if this wasn't in the negativeCache in the first place so it looks like this would always be a no-op anyhow

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/negativecache/NegativeCacheFacetImplTest.groovy
```diff
@@ -0,0 +1,278 @@
...
+  private NegativeCacheKey key
+  private Status status
+  private CacheManager cacheManager
+  private Ehcache cache
+  private Repository repository
+  private Map attributes
+
+  @Before
+  void setUp() {
+    cacheManager = mock(CacheManager)
```
kellyrob99: @adreghiciu Is it possible to replace the mock() calls with @Mock annotations? Cleans up the boilerplate quite a bit if so, and TestSupport already calls initMocks so it should "just work"

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/negativecache/NegativeCacheFacetImplTest.groovy
```diff
@@ -0,0 +1,278 @@
...
+    repository = mock(Repository)
+    when(repository.name).thenReturn('test')
+    Configuration config = mock(Configuration)
+    when(repository.configuration).thenReturn(config)
+    when(config.attributes(NegativeCacheFacetImpl.CONFIG_KEY)).thenReturn(
+        new NestedAttributesMap(NegativeCacheFacetImpl.CONFIG_KEY, attributes = Maps.newHashMap())
+    )
+  }
+
+  /**
```
kellyrob99: It's a shame that Spock and Mockito don't play very nicely together, as the given/then style is very nice in Spock.

kellyrob99:
+1

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/negativecache/NegativeCacheFacetImpl.java
adreghiciu: If negative cache is not not enabled in config, the cache is not created and such is null

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/negativecache/NegativeCacheHandler.java
adreghiciu: in case of concurrent access, one thread could result in 404 and such put it cache and another thread could get 200 and such it will invalidate the results of the other thread

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/negativecache/NegativeCacheFacetImplTest.groovy
adreghiciu: Sure, I can do that (still there will not be less lines of code :)

adreghiciu:
Manually merged into CMA3 https://github.com/sonatype/nexus-oss/commit/dcc0bdd68d6624746f9ff4eb2301333ceee3da8e

PR #1053, opened 2015-02-21
[NEXUS-8177][CMA] Repository availability (online/offline)

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8177


mrprescott:
+1. I agree it's faintly weird that the property is on viewFacet rather than on respository; I assume that's because repositories don't have events to pick up their new configuration?

adreghiciu:
We discussed with Jason that the property eventually will have to go on repository itself yet we can do that later on

plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/RepositoryCMAComponent.groovy
```diff
@@ -103,6 +104,7 @@ extends DirectComponentSupport
         name: input.name,
         type: input.type,
         format: input.format,
+        online: input.facet(ViewFacet).online,
```
kellyrob99: are we guaranteed to have a ViewFacet here? If there's doubt, might want to use ?. as a guard

kellyrob99:
+1

plugins/rapture/nexus-coreui-plugin/src/main/java/org/sonatype/nexus/coreui/RepositoryCMAComponent.groovy
adreghiciu: ViewFacet is always available. facet() call will throw an exception if facet is not available

adreghiciu:
Manually merged into CMA3 https://github.com/sonatype/nexus-oss/commit/ecdca3d6fab6e8b1710231bb73aa1acbfe2095dc

PR #1055, opened 2015-02-23
Refactors legacy repository UI into a separate rapture plugin

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8007

Related PRs:

https://github.com/sonatype/nexus-pro/pull/491
https://github.com/sonatype/nexus-bundles/pull/131

The old repository views have been put in their own plugin and marked as legacy in the UI. Repository-related strings have been prefixed with `LEGACY_`.

repositories - sonatype nexus 2015-02-25 09-40-01


jdillon:
Probably do not want to move all the resources (icons), probably not worth trying to move the icons used by legacy either, since those will be re-used most likely for the new ui bits.

djsauble:
Open question to @jdillon, @adreghiciu, and @kellyrob99: what other parts of the UI should be refactored into legacy? Staging rules/profiles, health check, browse repositories, and upload artifacts all come to mind.

jdillon:
@djsauble for now just the things which conflict with what we are rebuilding.  And really if its causing grief we may just remove those for now and rebuild them.

adreghiciu:
+1

PR #1062, opened 2015-02-26
[NEXUS-8195][2.11.x] Deadlock between yum merge metadata task and yum generate metadata task

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8195

Move marking of a Yum group repository as dirty outside of block that we keep the UID lock to avoid deadlock between the group that has to be marked and its member repository.


plugins/yum/nexus-yum-repository-plugin/src/main/java/org/sonatype/nexus/yum/internal/task/GenerateMetadataTask.java
```diff
@@ -165,27 +165,26 @@ protected YumRepository doRun()
...
-
-      regenerateMetadataForGroups();
-      return new YumRepositoryImpl(getRepoDir(), repositoryId, getVersion());
     }
     finally {
       mdUid.getLock().unlock();
     }
+
+    // TODO dubious
+    Thread.sleep(100);
```
kellyrob99: No less dubious now  but I trust you anyway ;)

kellyrob99:
+1

cstamas:
+1

PR #1063, opened 2015-02-27
[NEXUS-8014][CMA] Repository remote connection status/availability

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8014

Remote availability is checked at http client level. This ensures that availability check is done / status is updated regardless from where remote is accessed(e.g.ProxyFacetSupport, NugetFeedFetcher), and without interfering with the code that makes the access.


components/nexus-repository/src/main/java/org/sonatype/nexus/repository/httpclient/RemoteConnectionStatus.java
```diff
@@ -0,0 +1,52 @@
...
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.repository.httpclient;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * @since 3.0
+ */
+public class RemoteConnectionStatus
```
cstamas: Don't we need some `isAvailable` boolean method on this? As currently only description/reason differentiates statuses, that I guess are mostly for UI...

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/httpclient/FilteredHttpClient.java
```diff
@@ -0,0 +1,354 @@
...
+          synchronized (this) {
+            // avoid some other thread already increased the sequence
+            if (blockedUntil == null || blockedUntil.isBeforeNow()) {
+              blockedUntil = DateTime.now().plus(autoBlockSequence.next());
+              if (checkThread != null) {
+                checkThread.interrupt();
+              }
+              String uri = target.toURI();
+              // TODO maybe find different means to schedule status checking
+              checkThread = new Thread(new CheckStatus(uri, blockedUntil), "Check Status " + uri);
```
cstamas: In NX2 repositories had a thread pool for tasks like these. This not only helped centralising the work, but also helps a lot wrt thread naming (to be repo specific) and makes better logging, easier to connect status checking thread with repo being checked.

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/httpclient/FilteredHttpClient.java
```diff
@@ -0,0 +1,354 @@
...
+    blocked = checkNotNull(config).getConnectionConfig().isBlocked();
+    autoBlock = config.getConnectionConfig().shouldAutoBlock();
+    status = new RemoteConnectionStatus(blocked ? "Remote Manually Blocked" : "Unknown");
+    // TODO shall we use config.getConnectionConfig().getTimeout() * 2 as in NX2?
+    autoBlockSequence = new FibonacciNumberSequence(Time.seconds(40).toMillis());
+  }
+
+  private <T> T filter(final HttpHost target, final Filterable<T> filterable) throws IOException {
+    // main target is the first accessed target
+    if (mainTarget == null) {
```
cstamas: Concurrent calls might end up with wrong mainTarget set?

adreghiciu: I did not want to sync on this as in my view different threads for same repo they all start with same target so regardless the task they should all end up with same main target

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/httpclient/FilteredHttpClient.java
adreghiciu: Discussed about this with Jason and we postponed having a thread pool till we have a better infrastructure in place. Put the TODO there so we revisit

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/httpclient/RemoteConnectionStatus.java
adreghiciu: There is no use case for it yet so I did not add it

kellyrob99:
+1

cstamas:
+1

adreghiciu:
Manually merged into nexus-cma3 https://github.com/sonatype/nexus-oss/commit/3d31fc0a30bcb967085afcf9bc0310143fb99f42

PR #1066, opened 2015-03-02
[CMA] Index components via Orient DB hooks

adreghiciu:
Moving to a hook it will ensure that indexing of components will always be run and eliminate the need for repository implementation to call into indexing


kellyrob99:
+1

cwilper:
+1

mrprescott:
+1 I like it, I like that it's low level so it can't miss anything, but it's controlled by storage. It's okay that we're indexing assets as well as components?

cstamas:
+1

adreghiciu:
@mrprescott we are only index on component type events, so we do nothing when assets change, which may be problematic if we update some asset without changing the component in any way

mrprescott:
@adreghiciu Got it. I missed the filtering on V_COMPONENT. We may need to revisit for maven.

jdillon:
Curious, why?  Was previous means/api to manage the index not working?

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/StorageFacetImpl.java
```diff
@@ -85,6 +96,16 @@ protected void doInit() throws Exception {
     super.doInit();
   }
 
+  @Override
+  protected void doStart() throws Exception {
+    try {
+      searchFacet = getRepository().facet(SearchFacet.class);
+    }
+    catch (MissingFacetException e) {
+      // no search facet, no indexing
```
jdillon: are there real cases where a repo would not have a search facet?  ie is the nullability here causing additional logic/code to manage a situation that likely will never happen?

adreghiciu:
Indexing was all working fine, but you have to call into indexing facet in different places.
Moving this to a hook will ensure that indexing is always done regardless from where the components are changed.
As we also figured out you have to call after transaction is committed as the RID we used is not set if transaction is not committed.

adreghiciu:
Manually merged into nexus-cma3 https://github.com/sonatype/nexus-oss/commit/4e589cb344cadf0267bbd42b7b721807ddc7b3ac

PR #1068, opened 2015-03-03
Fixes load order of JS files when in production mode

djsauble:
Forces `nexus-coreui_legacy-plugin-prod.js` to load before `nexus-coreui-plugin-prod.js`, which eliminates dependency errors when running the console without the `?debug` flag.


kellyrob99:
+1

mrprescott:
+1

PR #1074, opened 2015-03-04
Fixes sort algorithm so breadcrumb truncation will work in IE

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8203

Sort function expects return value of 1, 0, and -1 when `a` is less than, equal to, or greater than `b`, respectively. I had been returning true/false instead, which somehow worked in Chrome/Firefox/Safari, but broke IE.


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/drilldown/Drilldown.js
```diff
@@ -492,17 +492,24 @@ Ext.define('NX.view.drilldown.Drilldown', {
...
    */
   reduceButtonWidth: function(buttons, width, minPerButton) {
     var me = this,
       currentWidth = 0,
-      currentButtonWidth,
-      nextButtonWidth,
       setToWidth;
 
+    // Sort the buttons by width
+    buttons = buttons.sort(
```
kellyrob99: should be able to simplify to:
```return a.getWidth() - b.getWidth()```

kellyrob99:
+1

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/drilldown/Drilldown.js
djsauble: I’m a sucker for simplification. Amended.

djsauble:
Manual merge.

PR #1075, opened 2015-03-05
Adds period, revises connection rules text slightly

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-7758

Text revision based on Tuesday’s demo feedback. Clarifies that max retries include the first attempt.


kellyrob99:
+1, although a simple text change like this doesn't require a pull IMHO

PR #1078, opened 2015-03-05
TokenParser improvements

cstamas:
Enforce same named variable tokens have equal values.


components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/matchers/token/TokenParser.java
```diff
@@ -84,7 +85,16 @@ public TokenParser(final String templatePattern) {
 
     Map<String, String> values = new HashMap<>();
     for (int i = 0; i < matcher.groupCount(); i++) {
-      values.put(variables.get(i).getName(), matcher.group(i + 1));
+      if (values.containsKey(variables.get(i).getName())) {
+        if (!Objects.equals(values.get(variables.get(i).getName()), matcher.group(i + 1))) {
+          log.trace("Variable '{}' values mismatch: '{}' vs '{}'", variables.get(i).getName(),
+              values.get(variables.get(i).getName()), matcher.group(i + 1));
```
mrprescott: This change is a good idea, but now I regret not refactoring out a couple of local variables in my original code. I need more coffee to parse this. :)  What about:
    final String varName = variables.get(i).getName();
    final String value = matcher.group(i + 1);

mrprescott:
+1 with minor comment

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/matchers/token/TokenParser.java
jdillon: probably look at making this more readable... a few locals could really help make this readable, and avoid duplication of work done to log.trace (which asis should be guarded, but if the code was simplified likely wouldn't).

jdillon:
+1 much better, I can' not read this and my eyes don't bleed

adreghiciu:
+1

cstamas:
Merged as d3bf83090a41b4e23c33df0b786bd532c882b939

PR #1081, opened 2015-03-06
Set search breadcrumb through the feature, not the drilldown

djsauble:
@adreghiciu


jdillon:
seems reasonable

adreghiciu:
I merged to use the feature to set the icon.
Still will not solve name setting as onSelection will be called before drill down is setting the name based on getDescription(model). Th name will be set actually yet it will very shortly be overwritten by drill down.
I will implement getDescription and that should work

PR #1082, opened 2015-03-06
Maven CMA support

cstamas:
WIP


plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/Content.java
```diff
@@ -0,0 +1,35 @@
...
+public interface Content
+{
+  String getContentType();
+
+  long getSize();
+
+  InputStream openInputStream() throws IOException;
+
+  DateTime getLastUpdated();
+}
```
jdillon: This looks a lot like Payload, was this due to copying simple or raw?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/group/MavenGroupHandler.java
```diff
@@ -0,0 +1,152 @@
...
+import static org.sonatype.nexus.repository.http.HttpMethods.GET;
+import static org.sonatype.nexus.repository.maven.internal.ContextHelper.coordinates;
+import static org.sonatype.nexus.repository.maven.internal.ContextHelper.name;
+
+/**
+ * Maven group handler.
+ *
+ * @since 3.0
+ */
+// TODO: this is started from copy+paste of GroupHandler, but that class as-is cannot be extended to suit this use case
```
jdillon: What use case are you talking about?  We don't want to rebuild mirror copies of things if there is an issue with the underlying common bits like this if we can avoid it.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven1/Maven1Format.java
```diff
@@ -0,0 +1,35 @@
...
+import org.sonatype.nexus.repository.Format;
+
+/**
+ * Maven 1 legacy layout (used by Maven 1.x) repository format.
+ *
+ * @since 3.0
+ */
+@Named(Maven1Format.NAME)
+@Singleton
+public class Maven1Format
```
jdillon: Why are we concerned with maven1 now?  There is a seperate story for this

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2GroupRecipe.groovy
```diff
@@ -0,0 +1,116 @@
...
+    repository.attach(searchFacet.get());
+    repository.attach(mavenStorageFacet.get())
+    repository.attach(configure(viewFacet.get()))
+  }
+
+  private ViewFacet configure(final ConfigurableViewFacet facet) {
+    Router.Builder builder = new Router.Builder()
+
+    builder.route(new Route.Builder()
+        .matcher(new TokenMatcher("/{name:.+}"))
```
jdillon: Unsure why we aren't using the token matcher to match the m2 coordinates, that is what the token system is meant to do.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/storage/MavenStorageFacet.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.joda.time.DateTime;
+
+/**
+ * Provides persistent storage for {@link Content}.
+ *
+ * @since 3.0
+ */
+@Facet.Exposed
+public interface MavenStorageFacet
```
jdillon: Why do we need a custom storage facet interface here?  Or is the name just wonky, since this seems to have nothing to do with StorageFacet?

jdillon:
Keep the packaging simple, there is no reason why we need to organize packages like you have done so far at this time.  I recommend flattening all this down to 1-2 packages and a single internal package until it warrants packaging.

cstamas:
Will reopen PR against master, now that nexus-cma3 is merged into it.

PR #1083, opened 2015-03-06
[NEXUS-8212] Fixes error when rebuilding metadata

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8212

I was pointed at the wrong Ext.Direct endpoint, basically. Fixed.


djsauble:
Manual merge.

PR #1084, opened 2015-03-06
[NEXUS-8205] Adds solid background color for buttons in IE

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8205

ExtJS has a bug which ignores the background-gradient-* setting in IE10 and IE11, even though these browsers support gradients. I've opted to use solid colors as a workaround for now, since this bug should be resolved once we upgrade to ExtJS 5.

Before:
ie11 - win7 2015-03-06 10-11-47

After:
ie11 - win7 2015-03-06 10-03-21


adreghiciu:
+1

djsauble:
Manual merge. https://github.com/sonatype/nexus-oss/commit/dc97345e258d4c84d52cf3793de4551c3bfe464c

kellyrob99:
+1 and should probably link this on https://issues.sonatype.org/browse/NEXUS-7322

djsauble:
Linked.

PR #1085, opened 2015-03-06
Fixes a few problems with ellipses and copyright symbols on IE

djsauble:
Issues:
https://issues.sonatype.org/browse/NEXUS-8110
https://issues.sonatype.org/browse/NEXUS-8182

@joedragons doesn’t see the copyright symbol or ellipses in his copy of Windows/IE. I wonder if the problem is that I use literals in code, rather than the HTML shorthand… `©` instead of `&copy;`, for example.

This PR tests that theory on the copyright symbol in the footer.

welcome - sonatype nexus 2015-03-06 11-24-15


mrprescott:
+1

adreghiciu:
+1

PR #1086, opened 2015-03-07
NEXUS-7684: Maven Hosted Repository

cstamas:
Maven Hosted CMA support

Issue
https://issues.sonatype.org/browse/NEXUS-7684
https://issues.sonatype.org/browse/NEXUS-8028


plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/policy/RedeployPolicyHandler.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Handler enforcing redeploy policy on Maven repositories.
+ *
+ * @since 3.0
+ */
+@Singleton
+@Named
+public class RedeployPolicyHandler
+    extends ComponentSupport
+    implements Handler
```
jdillon: This same some general implications to the rest of the CMA framework:

https://issues.sonatype.org/browse/NEXUS-8015

A handler just to do this seems like overkill, but should instead be something wrapped up into the storage-facet I think.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/policy/VersionPolicyHandler.java
```diff
@@ -0,0 +1,57 @@
...
+    if (artifactCoordinates != null) {
+      final MavenFacet mavenFacet = context.getRepository().facet(MavenFacet.class);
+      final VersionPolicy versionPolicy = mavenFacet.getVersionPolicy();
+      final boolean isSnapshot = mavenFacet.getPathParser().isSnapshotVersion(artifactCoordinates.getVersion());
+      if ((SNAPSHOT.equals(versionPolicy) && !isSnapshot) || (RELEASE.equals(versionPolicy) && isSnapshot)) {
+        return HttpResponses.badRequest(
+            "VersionPolicy '" + versionPolicy + "' forbids version '" + artifactCoordinates.getVersion() + "'");
+      }
+    }
+    return context.proceed();
```
jdillon: Initial gut feeling is that we don't need fine grained handlers like this, but perhaps instead this can/should be rolled into a maven-format specific handler to deal with these maven-specifics?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/PathParser.java
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Parses Maven path into {@link Coordinates}.
+   */
+  @Nonnull
+  Coordinates parsePath(String path);
+
+  /**
+   * Returns {@code true} if version string represents a Maven SNAPSHOT version.
+   */
+  boolean isSnapshotVersion(String version);
```
jdillon: PathParser and friends, seem like they should probably be a maven-specific Matcher, and/or refinements to the token-matcher to cope with the Maven format requirements around matching coordinates.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/ContentHelper.java
```diff
@@ -0,0 +1,69 @@
...
+import org.sonatype.nexus.repository.view.payloads.StreamPayload;
+
+import org.joda.time.DateTime;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * @since 3.0
+ */
+public class ContentHelper
```
jdillon: This seems qustionable and looks similar to other view payload helpers.

I think its fine for wip/investigation, be we probably don't want each format needing to redefine these things.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/ContextHelper.java
```diff
@@ -0,0 +1,59 @@
...
+import org.sonatype.nexus.repository.view.Context;
+import org.sonatype.nexus.repository.view.matchers.token.TokenMatcher;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * @since 3.0
+ */
+public class ContextHelper
```
jdillon: As mentioned elsewhere, if we have a mavne-specific matcher, then I think this and the other bits related become uneeded.

FTR the matcher was meant to perform the "path" -> coordinates used by remaining handlers bits, so by not using that we are building in some complexity.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacet.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @since 3.0
+ */
+@Facet.Exposed
+public interface MavenFacet
+    extends Facet
+{
+  VersionPolicy getVersionPolicy();
+
+  RedeployPolicy getRedeployPolicy();
```
jdillon: As mentioned elsewhere, redeployment policy seems like its a general aspect to all repos not specific to maven.  Ie. any repository could allow redeployment or only allow a single deployment?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenGroupHandler.java
```diff
@@ -0,0 +1,149 @@
...
+ * Maven group handler.
+ *
+ * @since 3.0
+ */
+// TODO: this is started from copy+paste of GroupHandler, but that class as-is cannot be extended to suit this use case
+@Singleton
+@Named
+public class MavenGroupHandler
+    extends ComponentSupport
+    implements Handler
```
jdillon: FTR, lets focus on one type at a time.  So lets get hosted working first.  Then proxy, then we can sort out group.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2IndexSettingsContributor.java
```diff
@@ -0,0 +1,47 @@
...
+public class Maven2IndexSettingsContributor
+    implements IndexSettingsContributor
+{
+
+  @Override
+  public String getIndexSettings(final Repository repository) throws IOException {
+    if (Maven2Format.NAME.equals(repository.getFormat().getValue())) {
+      return Resources
+          .toString(Resources.getResource(Maven2IndexSettingsContributor.class, "es-mapping-maven.json"), UTF_8);
+    }
```
jdillon: I've not yet looked, but if we have this for each format, this screams out for a tiny support class to hide these details.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/storage/MavenContentsFacet.java
```diff
@@ -0,0 +1,45 @@
...
+ * @since 3.0
+ */
+@Facet.Exposed
+public interface MavenContentsFacet
+    extends Facet
+{
+  @Nullable
+  Content get(Coordinates coordinates) throws IOException;
+
+  void put(Coordinates coordinates, Content content) throws IOException, InvalidContentException;
```
jdillon: InvalidContextException, seems like it may be better off as a RuntimeException, as its dependent on input.  Or is there some indent that anything other than wrapping, rethrowing or ignoring of this exception is possible?

plugins/basic/nexus-repository-maven/src/test/java/org/sonatype/nexus/repository/maven/internal/ArtifactCoordinatesTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
+
+public class ArtifactCoordinatesTest
```
jdillon: Try to get into the habbit of always putting class javadocs.  Even in this case of:

```
Tests for {@link ArtifactCoordinates}.
```

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenProxyFacet.java
```diff
@@ -0,0 +1,93 @@
...
+  public static final String CONFIG_KEY = "maven";
+
+  private ChecksumPolicy checksumPolicy;
+
+  @Override
+  protected void doConfigure() throws Exception {
+    super.doConfigure();
+    NestedAttributesMap attributes = getRepository().getConfiguration().attributes(CONFIG_KEY);
+    checksumPolicy = ChecksumPolicy.valueOf(
+        attributes.get("checksumPolicy", String.class, ChecksumPolicy.WARN.getValue())
```
mrprescott: Should this be attributes.require, since .valueOf will NPE otherwise?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenProxyFacet.java
```diff
@@ -0,0 +1,93 @@
...
+    );
+  }
+
+  @Override
+  protected Payload getCachedPayload(final Context context) throws IOException {
+    final MavenContentsFacet contentsFacet = getRepository().facet(MavenContentsFacet.class);
+    if (context.getAttributes().contains(ArtifactCoordinates.class)) {
```
mrprescott: Calling .contains() then .require is amusing. What about just:
    final ArtifactCoordinates artifactCoords = context.getAttributes().get(ArtifactCoordinates.class);
    if(artifactCoords != null) {
      return contentsFacet.getArtifact(artifactCoords);
    }

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenProxyFacet.java
cstamas: It provides a default value for attributes.get... or do I misinterpret the API?

jdillon: ya, get with a default should never return null, unless the default was null.

btw if we start doing lots of this enum conversion probably should add convert support sooner rather than later, as the intent was do be able to do something like this:

```
checksumPolicy = attributes.get("checksumPolicy", ChecksumPolicy.class, ChecksumPolicy.WARN)
```

mrprescott: My bad. :)

plugins/basic/nexus-repository-maven/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>org.sonatype.nexus</groupId>
+      <artifactId>nexus-repository</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.sonatype.nexus</groupId>
+      <artifactId>nexus-elasticsearch</artifactId>
```
jdillon: this is already a dependency from nexus-repository, don't duplicate

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenArtifactHandler.java
```diff
@@ -0,0 +1,81 @@
...
+  public Response handle(final @Nonnull Context context) throws Exception {
+    final String action = context.getRequest().getAction();
+    final MavenFacet mavenFacet = context.getRepository().facet(MavenFacet.class);
+    final ArtifactCoordinates coordinates = context.getAttributes().require(ArtifactCoordinates.class);
+
+    final VersionPolicy versionPolicy = context.getRepository().facet(MavenFacet.class).getVersionPolicy();
+    if (!versionPolicy.allowsCoordinates(coordinates)) {
+      return HttpResponses.badRequest(
+          "Repository version policy '" + versionPolicy + "' does not allow version '" +
+              coordinates.getVersion() + "'");
```
jdillon: Try to avoid quoting in text strings as this just complicates the code w/o good reason.

```
HttpResponses.badRequest("Repository version policy: " + versionPolicy + " does not allow version: " + coordinates.getVersion());
```

String.format if you must

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenChecksumReader.java
```diff
@@ -0,0 +1,124 @@
...
+
+import javax.annotation.Nullable;
+
+import org.sonatype.nexus.repository.view.Payload;
+import org.sonatype.sisu.goodies.common.ComponentSupport;
+
+import com.google.common.base.Charsets;
+import com.google.common.base.Strings;
+import com.google.common.io.CharStreams;
+import org.codehaus.plexus.util.StringUtils;
```
jdillon: Do not use plexus anything

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenChecksumReader.java
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Utility class for reading up Maven .sha1/.md5 files that might be created by vast different tools out there.
+ *
+ * @since 3.0
+ */
+public class MavenChecksumReader
+    extends ComponentSupport
```
jdillon: Why do we need this?  Storage api should do this for assets stored.  Why do we need additional handling here?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenProxyFacet.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Maven proxy facet.
+ *
+ * @since 3.0
+ */
+@Named
+public class MavenProxyFacet
+    extends ProxyFacetSupport
+{
+  public static final String CONFIG_KEY = "maven";
```
jdillon: MavenFacet and MavenProxyFacet sharing sake config key?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/policy/ChecksumPolicy.java
```diff
@@ -0,0 +1,96 @@
...
+  private static final String V_STRICT = "strict";
+
+  public static final ChecksumPolicy IGNORE = new ChecksumPolicy(V_IGNORE);
+
+  public static final ChecksumPolicy WARN = new ChecksumPolicy(V_WARN);
+
+  public static final ChecksumPolicy STRICT_IF_EXISTS = new ChecksumPolicy(V_STRICT_IF_EXISTS);
+
+  public static final ChecksumPolicy STRICT = new ChecksumPolicy(V_STRICT);
+
```
jdillon: Why not an enum?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/policy/VersionPolicy.java
```diff
@@ -0,0 +1,102 @@
...
+  private static final String V_SNAPSHOT = "snapshot";
+
+  private static final String V_MIXED = "mixed";
+
+  public static final VersionPolicy RELEASE = new VersionPolicy(V_RELEASE);
+
+  public static final VersionPolicy SNAPSHOT = new VersionPolicy(V_SNAPSHOT);
+
+  public static final VersionPolicy MIXED = new VersionPolicy(V_MIXED);
+
```
jdillon: why not an enum?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/Coordinates.java
```diff
@@ -0,0 +1,184 @@
...
+import javax.annotation.concurrent.Immutable;
+
+import org.sonatype.nexus.common.hash.HashAlgorithm;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Maven non-artifact repository coordinates. Every item in repository may have hashes, stored on paths with proper
+ * suffixes.
```
jdillon: I'm not sure this represents a "coordinate" system, its just a path.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenArtifactMatcher.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public class MavenArtifactMatcher
+    extends ComponentSupport
+    implements Matcher
+{
+  @Override
+  public boolean matches(final Context context) {
+    final String path = context.getRequest().getPath();
+    final MavenFacet mavenFacet = context.getRepository().facet(MavenFacet.class);
+    final ArtifactCoordinates coordinates = mavenFacet.getArtifactCoordinatesParser().parsePath(path);
```
jdillon: Why does matcher need facet here?

jdillon:
Recommend you remove group/proxy bits until we have hosted sorted as it just inflates the review space and perceived complexity of the change you are adding to introduce hosted support.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenArtifactHandler.java
cstamas: Ack.

Maybe add some template support for the `HttpResponses` method instead? String.format is rather expensive (by factors of 5-6 more than plain + for example).

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenChecksumReader.java
cstamas: There is a TODO to remove it.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenChecksumReader.java
cstamas: This class is needed to "interpret" checksum files coming from "outside", from either clients upload (deploying them) or from proxied remotes. In both cases, we did saw very strange files, and this class accumulated all the knowledge from NX2 and is able to interpret .sha1/.md5 files in same way as Maven does. Once this class "extracted" hash, is not used anymore, as hash will come from asset attributes or asset itself.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/Coordinates.java
cstamas: Ack. Will rename it to something like RepositoryPath. The point of the class is to ease interpreting maven repository paths, where /foo or /foo.sha1 or /foo.md5 all actually refers to /foo or some of it's "subordinate" properties like SHA1 or MD5. Also, it simplifies path checking, as removes the need to consider optionally present extensions:

`coordinates.main().getFileName().equals(MAVEN_METADATA_FILENAME)`

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenArtifactMatcher.java
cstamas: Am tinkering between the two: have ArtifactCoordinatesParser exposed via facet or not. Originally it was not, and matcher had it injected via ctor, then I exposed it via facet. Will undo, as currently only matcher needs parser, nothing else, no need to have it on a facet (although am unsure about the future).

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenArtifactHandler.java
jdillon: Na, just use string concat if you worry about String.format, its sending a response, in this case will render an error template which is even more expensive than format.

My point was don’t bother adding quotes like this which makes for very ugly code just to put quotes around things.  You also end up putting quotes around everything and then it starts to land-slide in complexity w/o good reason… that was my point.

—jason




On March 11, 2015 at 1:29:55 AM, Tamas Cservenak (notifications@github.com) wrote:

In plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenArtifactHandler.java:

> +public class MavenArtifactHandler
> +    extends ComponentSupport
> +    implements Handler
> +{
> +  @Nonnull
> +  @Override
> +  public Response handle(final @Nonnull Context context) throws Exception {
> +    final String action = context.getRequest().getAction();
> +    final MavenFacet mavenFacet = context.getRepository().facet(MavenFacet.class);
> +    final ArtifactCoordinates coordinates = context.getAttributes().require(ArtifactCoordinates.class);
> +
> +    final VersionPolicy versionPolicy = context.getRepository().facet(MavenFacet.class).getVersionPolicy();
> +    if (!versionPolicy.allowsCoordinates(coordinates)) {
> +      return HttpResponses.badRequest(
> +          "Repository version policy '" + versionPolicy + "' does not allow version '" +
> +              coordinates.getVersion() + "'");
Ack.

Maybe add some template support for the HttpResponses method instead? String.format is rather expensive (by factors of 5-6 more than plain + for example).

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenChecksumReader.java
jdillon: TODO have a habbit of never getting TODONE, so where you can avoid them by simply not needing a TODO in this case using another non-plexus String helper like Strings from guava or String2 from nexus-common, use it.

—jason




On March 11, 2015 at 1:30:09 AM, Tamas Cservenak (notifications@github.com) wrote:

In plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenChecksumReader.java:

> + */
> +package org.sonatype.nexus.repository.maven.internal;
> +
> +import java.io.IOException;
> +import java.io.InputStreamReader;
> +import java.util.Locale;
> +
> +import javax.annotation.Nullable;
> +
> +import org.sonatype.nexus.repository.view.Payload;
> +import org.sonatype.sisu.goodies.common.ComponentSupport;
> +
> +import com.google.common.base.Charsets;
> +import com.google.common.base.Strings;
> +import com.google.common.io.CharStreams;
> +import org.codehaus.plexus.util.StringUtils;
There is a TODO to remove it.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenChecksumReader.java
jdillon: Why does it need to interpret them?

—jason




On March 11, 2015 at 1:32:25 AM, Tamas Cservenak (notifications@github.com) wrote:

In plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenChecksumReader.java:

> +
> +import org.sonatype.nexus.repository.view.Payload;
> +import org.sonatype.sisu.goodies.common.ComponentSupport;
> +
> +import com.google.common.base.Charsets;
> +import com.google.common.base.Strings;
> +import com.google.common.io.CharStreams;
> +import org.codehaus.plexus.util.StringUtils;
> +
> +/**
> + * Utility class for reading up Maven .sha1/.md5 files that might be created by vast different tools out there.
> + *
> + * @since 3.0
> + */
> +public class MavenChecksumReader
> +    extends ComponentSupport
This class is needed to "interpret" checksum files coming from "outside", from either clients upload (deploying them) or from proxied remotes. In both cases, we did saw very strange files, and this class accumulated all the knowledge from NX2 and is able to interpret .sha1/.md5 files in same way as Maven does. Once this class "extracted" hash, is not used anymore, as hash will come from asset attributes or asset itself.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenArtifactMatcher.java
jdillon: While its technically possible for matchers to use facets I never imagined the matchers were to be complex enough to ever need that.  And it could present some coupling issues, so I’d avoid it where possible/feasible and leave the matchers as simple token-like data extraction components.

—jason




On March 11, 2015 at 1:36:40 AM, Tamas Cservenak (notifications@github.com) wrote:

In plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenArtifactMatcher.java:

> +import org.sonatype.sisu.goodies.common.ComponentSupport;
> +
> +/**
> + * Matcher that matches for Maven artifacts only, and sets {@link ArtifactCoordinates} in context attributes.
> + *
> + * @since 3.0
> + */
> +public class MavenArtifactMatcher
> +    extends ComponentSupport
> +    implements Matcher
> +{
> +  @Override
> +  public boolean matches(final Context context) {
> +    final String path = context.getRequest().getPath();
> +    final MavenFacet mavenFacet = context.getRepository().facet(MavenFacet.class);
> +    final ArtifactCoordinates coordinates = mavenFacet.getArtifactCoordinatesParser().parsePath(path);
Am tinkering between the two: have ArtifactCoordinatesParser exposed via facet or not. Originally it was not, and matcher had it injected via ctor, then I exposed it via facet. Will undo, as currently only matcher needs parser, nothing else, no need to have it on a facet (although am unsure about the future).

cstamas:
Just verified hosted repo by locally building and deploying https://github.com/apache/maven to it. Also, tested some retrieval operations, all went well. For setup details see [README](https://github.com/sonatype/nexus-oss/blob/nexus-8028-maven-ng/plugins/basic/nexus-repository-maven/README.md)

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,614 @@
...
+      assetAttributes.set(P_ASSET_KEY, getAssetKey(path));
+      assetAttributes.set(P_ASSET_KIND, ASSET_KIND_ARTIFACT);
+      assetAttributes.set(P_GROUP_ID, path.getGroupId());
+      assetAttributes.set(P_ARTIFACT_ID, path.getArtifactId());
+      assetAttributes.set(P_VERSION, path.getVersion());
+      if (path.isSnapshot()) {
+        assetAttributes.set(P_BASE_VERSION, path.getBaseVersion());
+      }
+      if (path.getClassifier() != null) {
+        assetAttributes.set(P_CLASSIFIER, path.getClassifier());
```
jdillon: avoid guarding to set null, null value is allowed on attributes

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,614 @@
...
+              new MavenPayload(
+                  new StringPayload(hashCode.toString(), "text/plain"),
+                  content.getLastUpdated(),
+                  content.getLastModified(),
+                  Collections.<HashAlgorithm, HashCode>emptyMap()
+              )
+          );
+        }
+      }
+    }
```
jdillon: What is this doing?  automatically making hash assets?  Why?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenPayload.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  // ==
+
+  public static Response ok(final MavenPayload payload) {
+    final Response response = HttpResponses.ok(payload);
+    final HashCode sha1 = payload.getHashCode(HashAlgorithm.SHA1);
+    checkArgument(sha1 != null); // sanity
+    // TODO: this clearly creeps HTTP protocol in here, Date formatting and headers
+    response.getHeaders().set(HttpHeaders.ETAG, "{SHA1{" + sha1.toString() + "}}");
```
jdillon: Do not .toString() an object when its being concatenated with a string constant, this is pointless and verbose w/o reason.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/ArtifactPath.java
```diff
@@ -0,0 +1,298 @@
...
+
+  /**
+   * Returns {@code true} if this path is subordinate of another path.
+   *
+   * @see {@link #subordinateOf()}
+   */
+  @Override
+  public boolean isSubordinate() {
+    return isSignature() || super.isSubordinate();
+  }
```
jdillon: I question the value of this, why do we care if another path is subordinate or not?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/DigestExtractor.java
```diff
@@ -0,0 +1,159 @@
...
+  @Nullable
+  public static String extract(final InputStream stream)
+      throws IOException
+  {
+    checkNotNull(stream);
+    try (InputStreamReader isr = new InputStreamReader(
+        new LimitedInputStream(stream, 0, MAX_CHARS_NEEDED), Charsets.UTF_8)) {
+      return extract(CharStreams.toString(isr));
+    }
+  }
```
jdillon: Why do we need public api for any of this?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/HostedHandlerSupport.java
```diff
@@ -0,0 +1,106 @@
...
+import static org.sonatype.nexus.repository.http.HttpMethods.GET;
+import static org.sonatype.nexus.repository.http.HttpMethods.PUT;
+import static org.sonatype.nexus.repository.maven.internal.MavenPayload.ok;
+
+/**
+ * Maven hosted handler base.
+ *
+ * @since 3.0
+ */
+public abstract class HostedHandlerSupport<C extends Path>
```
jdillon: Generics here could be an indication of over complexity

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/HostedHandlerSupport.java
```diff
@@ -0,0 +1,106 @@
...
+    extends ComponentSupport
+    implements Handler
+{
+  private final Clock clock = new Clock();
```
jdillon: FTR the point of Clock was to allow it to be mocked out.  So using a Clock in a private final is kinda fucking pointless.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/HostedHandlerSupport.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+        return HttpResponses.noContent();
+      }
+
+      default:
+        return HttpResponses.methodNotAllowed(context.getRequest().getAction(), GET, PUT, DELETE);
+    }
+  }
+
+  protected abstract C getPath(final Context context) throws IllegalArgumentException;
```
jdillon: IllegalArgumentException is runtime, no reason to declare this

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/DigestExtractor.java
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * Returns {@code true} if the input string "looks like" digest hex (md5 or sha1).
+   */
+  public static boolean isDigest(final String digest) {
+    return digest.length() >= 32 && digest.matches("^[a-z0-9]+$");
+  }
+
+  // ==
```
jdillon: Please stop putting in these markers.  If you want to denote sections, then give them details.  I've found so many places where code is littered it "// ==" markers even ever other method in some places.  Its meaningless noise.

I don't object to making logical sections in code to organize related things, but name them and keep them updated.  With "//  ==" having no real meaning or context its var to easy for the organization you are indending to degrade very  quickly.

plugins/basic/nexus-repository-maven/src/test/java/org/sonatype/nexus/repository/maven/internal/DigestExtractorTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+  private InputStream stream(String string)
+      throws IOException
+  {
+    return new ByteArrayInputStream(string.getBytes("UTF-8"));
+  }
+
+  @Test
+  public void testAcceptedDigests()
+      throws IOException
```
jdillon: minor, just throw Exception and mind the formatting this is old style

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2Format.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.sonatype.nexus.repository.maven.internal.maven2;
+
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.sonatype.nexus.repository.Format;
+
+/**
+ * Maven 2 aka "default" layout (since Maven 2.x) repository format.
```
jdillon: Minor lets not mix terms.  Maven 2 format. should be sufficent for the lexicon of CMA but "default" and "layout" is a Maven term.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenPayload.java
```diff
@@ -0,0 +1,133 @@
...
+import static org.sonatype.nexus.common.hash.HashAlgorithm.MD5;
+import static org.sonatype.nexus.common.hash.HashAlgorithm.SHA1;
+
+/**
+ * Maven specific payload that enriches {@link Payload}.
+ *
+ * @since 3.0
+ */
+public class MavenPayload
+    implements Payload
```
jdillon: Don't know why we need a custom format specific Payload.  If there is information missing for the format, I think there are probably better/simpler means to express this other than sub-classes of otherwise pretty simple byte[]/type container

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2ArtifactPathParser.java
```diff
@@ -0,0 +1,82 @@
...
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.repository.maven.internal.maven2;
+
+import org.sonatype.nexus.proxy.maven.gav.Gav;
+import org.sonatype.nexus.proxy.maven.gav.M2GavCalculator;
```
jdillon: Wow... DO NOT DEPEND ON ANY NEXUS 2 CODE HERE

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,614 @@
...
+
+  private static final String ASSET_KIND_METADATA = "metadata";
+
+  // ==
+
+  private final MimeSupport mimeSupport;
+
+  public static final String CONFIG_KEY = "maven";
+
+  // ==
```
jdillon: here and above pointless "// ==" markers, arent' clearly providing any organization value, only noise

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/payloads/BlobPayload.java
```diff
@@ -0,0 +1,69 @@
...
+  private boolean opened = false;
+
+  public BlobPayload(final Blob blob, final @Nullable String contentType)
+  {
+    this.blob = checkNotNull(blob);
+    this.contentType = contentType;
+  }
+
+  @Override
+  public synchronized InputStream openInputStream() throws IOException {
```
cwilper: Why synchronized?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
cstamas: This code executes only when `!content.getHashAlgorithms().isEmpty()`, and they are always empty in case of hosted repository. The hashes on payload are set on proxy repository (other PR), where client fetches content AND hashes preemptively, before client even asked for them.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
cstamas: ack

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenPayload.java
cstamas: ack

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/DigestExtractor.java
cstamas: This class is in internal package (like everything else of maven stuff so far is), is not public API. Will lessen class visibility if that's enough.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/HostedHandlerSupport.java
cstamas: ack

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/DigestExtractor.java
cstamas: ack

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2ArtifactPathParser.java
cstamas: This is internal only to Maven2ArtifactPathParser internals, and used it to move faster. There is a TODO to move/update/flesh out stuff needed for GAV parsing from legacy code.

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/payloads/BlobPayload.java
cstamas: c+p, fixing

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/DigestExtractor.java
jdillon: If you actually need all these helpers, then leave them, but it seems like its the start of complexity to expose all the various byte-container parameters.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
jdillon: What is the requirement for this?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2ArtifactPathParser.java
jdillon: TODO have a very bad habit of never getting TODONE, so you are better off to avoid them from the start where possible and certainly anything that pulls in legacy should be avoided in CMA format impls.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/HostedHandlerSupport.java
cstamas: The truth is that I am not quite satisfied with these too: the reason of this stems from separate routes for artifacts and metadata. W/o HostedHandlerSupport the two handler would be 90% same, repetitive.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/HostedHandlerSupport.java
cstamas: Ack, I just like to have it written, making me remember to handle it.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/HostedHandlerSupport.java
jdillon: I recommend avoiding generics, except where isolated in helpers or where there is a clear pattern needed where generics simplifies things. More often than not it adds complexity instead of reducing it... so please keep that in mind.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/HostedHandlerSupport.java
jdillon: You should not _need_ to handle a RuntimeException that is part of the entire concept related to that classification of exception.  If you want a reminder add javadoc.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,426 @@
...
+  {
+    final String componentKeyName =
+        StorageFacet.P_ATTRIBUTES + "." + getRepository().getFormat().getValue() + "." + P_COMPONENT_KEY;
+    return tx.findComponentWithProperty(componentKeyName, getComponentKey(coordinates), bucket);
+  }
+
+  /**
+   * Returns a list of all component assets.
+   */
+  private List<OrientVertex> getComponentAssets(final OrientVertex component)
```
cwilper: Just a note, I'll probably throw this on the upcoming Component wrapper class for NEXUS-8278, and it will return Assets.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,426 @@
...
+
+  private NestedAttributesMap getFormatAttributes(final StorageTx tx, final OrientVertex vertex) {
+    return getAttributes(tx, vertex).child(getRepository().getFormat().getValue());
+  }
+
+  /**
+   * Determines or confirms the content type for the content, or throws {@link InvalidContentException} if it cannot.
+   */
+  @Nonnull
+  private String determineContentType(final MavenPath mavenPath,
```
cwilper: Looks very similar to RawContentFacetImpl#determineContentType, just takes a different kind of path. Seems like a common utility method somewhere would be handy.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,426 @@
...
+          try (TempStreamSupplier supplier = new TempStreamSupplier(inputStream)) {
+            final String contentType = determineContentType(path, supplier, content.getContentType());
+            try (InputStream is = supplier.get()) {
+              tx.setBlob(is, headers, asset, HASH_ALGORITHMS, contentType);
+            }
+          }
+        }
+        tx.commit();
+      }
+      catch (IOException | InvalidContentException e) {
```
cwilper: Calling tx.rollback explicitly is unnecessary, so this catch block can just be dropped: StorageTx automatically rolls back the underlying orient transaction at close time. So if the code in the block above didn't reach commit(), things will actually be rolled back. If it did reach commit(), the auto-rollback at close time will have no effect (the prior transaction is already committed)

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
cstamas: Yup, it does originate from raw, and basically MavenPath type is irrelevant, all it needs is is actually request path (same as in raw) so it does makes sense to pull out.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
cstamas: Ack

plugins/basic/nexus-repository-httpbridge/src/main/java/org/sonatype/nexus/repository/httpbridge/HttpResponseSender.java
```diff
@@ -26,5 +27,5 @@
  */
 public interface HttpResponseSender
 {
-  void send(Response response, HttpServletResponse httpServletResponse) throws ServletException, IOException;
+  void send(Response response, HttpServletRequest httpRequest, HttpServletResponse httpServletResponse) throws ServletException, IOException;
```
jdillon: Hrm.... unsure we want to use HttpServletRequest here

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,420 @@
...
+
+  private static final String CONFIG_KEY = "maven";
+
+  private static final List<HashAlgorithm> HASH_ALGORITHMS = Lists.newArrayList(SHA1, MD5);
+
+  // members
+
+  private final MimeSupport mimeSupport;
+
+  private boolean strictContentTypeValidation;
```
jdillon: Hrm... this seems like a general feature of repositories and not maven specific

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,420 @@
...
+        mavenPath.getPath());
+    return BlobRef.parse(blobRefStr);
+  }
+
+  /**
+   * Creates a payload out of a Blob, for content originating from Blobs.
+   */
+  private Payload marshall(final StorageTx tx, final OrientVertex asset, final Blob blob) {
+    final String contentType = asset.getProperty(StorageFacet.P_CONTENT_TYPE);
+    return new BlobPayload(blob, contentType);
```
jdillon: hrm... I don't think marshall() is a good name, its technically not marshalling anything here.  Its just adapting a blob to a payload

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenPath.java
```diff
@@ -0,0 +1,377 @@
...
+      if (getCoordinates() != null) {
+        mainCoordinates = new Coordinates(
+            getCoordinates().isSnapshot(),
+            getCoordinates().getGroupId(),
+            getCoordinates().getArtifactId(),
+            getCoordinates().getVersion(),
+            getCoordinates().getBaseVersion(),
+            getCoordinates().getClassifier(),
+            getCoordinates().getExtension().substring(0, getCoordinates().getExtension().length() - hashSuffixLen),
+            getCoordinates().getSignatureType()
```
jdillon: um, why not assign a local to getCoordinates() instead of invoking this method all the time?  Compiler _could_ optimize this, but really given how many times you need to call getCoordinates() it seems to suggest a local to resolve an object that would be better.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenPath.java
```diff
@@ -0,0 +1,377 @@
...
+    return null;
+  }
+
+  /**
+   * Returns path of passed in hash type that is subordinate of this path. This path cannot be hash.
+   */
+  @Nonnull
+  public MavenPath hash(final HashType hashType) {
+    checkNotNull(hashType);
+    checkArgument(!isHash(), "This path is already a hash: " + this);
```
jdillon: do not use + syntax for precodnitions, use %s (goes for all uses you have for this in this PR)

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2HostedRecipe.groovy
```diff
@@ -0,0 +1,118 @@
...
+                     @Named(Maven2Format.NAME) final Format format)
+  {
+    super(type, format)
+  }
+
+  @Override
+  void apply(@Nonnull final Repository repository) throws Exception {
+    repository.attach(securityFacet.get())
+    repository.attach(storageFacet.get())
+    repository.attach(searchFacet.get());
```
jdillon: avoid ; in groovy

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MavenPathParser.java
```diff
@@ -0,0 +1,83 @@
...
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.repository.maven.internal.maven2;
+
+import org.sonatype.nexus.proxy.maven.gav.Gav;
+import org.sonatype.nexus.proxy.maven.gav.M2GavCalculator;
```
jdillon: copy/extract/whatever but do not depend on legacy

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MavenPathParser.java
```diff
@@ -0,0 +1,83 @@
...
+ * Maven 2 path parser.
+ *
+ * @since 3.0
+ */
+public class Maven2MavenPathParser
+    implements MavenPathParser
+{
+  // TODO: TokenParser not capable to do this (yet)
+  // TODO: this is legacy, but here only to make it work
+  // TODO: consider moving/updating/fleshing out it from nx2 codebase, or improve TokenParser
```
jdillon: copy the code into internal package of this plugin for now, simplifies removal of legacy until probably a more proper solution to fix token-parser is done

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
cstamas: Yes, @cwilper mentioned this too, AFAIU it will be pulled out by him?

plugins/basic/nexus-repository-httpbridge/src/main/java/org/sonatype/nexus/repository/httpbridge/HttpResponseSender.java
cstamas: Um, yes, this PR contains one unrelated change: HEAD support. And HttpResponseSender must know the HTTP method to support that. Will undo these changes and create a backlog issue.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MavenPathParser.java
cstamas: Reimplementation is in progress

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MavenPathParser.java
cstamas: In progress

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2HostedRecipe.groovy
cstamas: ack

plugins/basic/nexus-repository-httpbridge/src/main/java/org/sonatype/nexus/repository/httpbridge/HttpResponseSender.java
jdillon: FTR I think the handling of HEAD in the sender is fine for this now.  My comment above was only to say use the Request not the HttpServerRequest.  So I'd change ResponseSender api to send(Request,Response,HttpServletResponse) so that the sender impl has all the information from request and response to render the servlet response.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/DigestExtractor.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  /**
+   * Extract the checksum from a string. If result appears to be a digest (cannot be 100% sure), it is returned,
+   * otherwise {@code null}.
+   */
+  @Nullable
+  public static String extract(final String input) {
+    checkNotNull(input);
+    // we need most of MAX_CHARS_NEEDED characters from input
```
mrprescott: minor: I think this should read: // we need at most MAX_CHARS_NEEDED

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/payloads/BlobPayload.java
```diff
@@ -0,0 +1,67 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * {@link Blob} backed payload.
+ *
+ * @since 3.0
+ */
+public class BlobPayload
```
mrprescott: I was looking at changing NuGet to use this (since it does the same thing), but the equivalent there is just this:
new StreamPayload(blob.getInputStream(), blob.getMetrics().getContentSize(), contentType)

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,413 @@
...
+          assetAttributes.set(P_ARTIFACT_ID, path.getCoordinates().getArtifactId());
+          assetAttributes.set(P_VERSION, path.getCoordinates().getVersion());
+          if (path.getCoordinates().isSnapshot()) {
+            assetAttributes.set(P_BASE_VERSION, path.getCoordinates().getBaseVersion());
+          }
+          assetAttributes.set(P_CLASSIFIER, path.getCoordinates().getClassifier());
+          assetAttributes.set(P_EXTENSION, path.getCoordinates().getExtension());
+        }
+
+        // TODO: if subordinate asset (sha1/md5/asc), should we link it somehow to main asset?
```
mrprescott: What did you decide? :)

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,413 @@
...
+  public void put(final MavenPath path, final Payload content)
+      throws IOException, InvalidContentException
+  {
+    try (StorageTx tx = getStorage().openTx()) {
+      OrientVertex component = null;
+      OrientVertex asset;
+      if (path.getCoordinates() != null) {
+        component = findComponent(tx, path.getCoordinates(), tx.getBucket());
+        if (component == null) {
+          component = tx.createComponent(tx.getBucket());
```
mrprescott: Maybe 'extract method' here.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,413 @@
...
+          }
+        }
+
+        asset = selectComponentAsset(tx, component, path);
+      }
+      else {
+        asset = findAsset(tx, path, tx.getBucket());
+      }
+      if (asset == null) {
+        asset = tx.createAsset(tx.getBucket());
```
mrprescott: Same here, 'extract method' - keep the logic of the outer method a lot easier to see.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,413 @@
...
+      OrientVertex asset;
+      if (path.getCoordinates() != null) {
+        component = findComponent(tx, path.getCoordinates(), tx.getBucket());
+        if (component == null) {
+          component = tx.createComponent(tx.getBucket());
+
+          // Set normalized properties: format, group, and name
+          component.setProperty(StorageFacet.P_FORMAT, getRepository().getFormat().getValue());
+          component.setProperty(StorageFacet.P_PATH, path.getPath());
+          component.setProperty(StorageFacet.P_GROUP, path.getCoordinates().getGroupId());
```
mrprescott: Local var for coordinates?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenPath.java
```diff
@@ -0,0 +1,374 @@
...
+  public String getFileName() {
+    return fileName;
+  }
+
+  @Nullable
+  public HashType getHashType() {
+    return hashType;
+  }
+
+  @Nullable
```
mrprescott: Bears repeating in javadoc: artifact paths (or their hashes and signatures) have coordinates, but mere file-paths don't. This is used to bifurcate a lot of behavior in the maven facet. An isArtifactPath() method would be good, too.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -0,0 +1,413 @@
...
+
+  @Nonnull
+  @Override
+  public ChecksumPolicy getChecksumPolicy() {
+    return checksumPolicy;
+  }
+
+  @Nullable
+  @Override
+  public Payload get(final MavenPath path) throws IOException {
```
mrprescott: see email

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
mrprescott: Even so, in the meantime surely this method should be dropped in favor of tx.findAssets(component) no? What's with the filtering on instanceof OrientVertex, what other vertices can be linked by a 'part of component' edge?

cwilper: Agree with @mrprescott that it should use tx.findAssets(component) for now, unless my PR makes it in first :) Also, no need for the instanceof check.

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/payloads/BlobPayload.java
cstamas: No, is not. AFAIK, `blob.getInputStream()` will OPEN the blob stream, whereas this one will defer opening it, as it is not needed for example in case of HEAD response.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
cstamas: I left the question open still, but I lean to "no, we do not want to link".

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenPath.java
cstamas: Isn't `Kind.ARTIFACT == mavenPath.getKind()` what you actually look for?

I agree, that above, and `mavenPath.getCoordinates() != null` _mean the same_, and might be confusing or misleading.

cstamas: Oh and one more thing: mere file-paths does not have signatures, but they might have hashes.

cstamas: The Kind was introduced only to simplify matchers implementation, see [MavenArtifactMatcher](https://github.com/sonatype/nexus-oss/blob/556e5dc4b1adebcde278e14f30f187a023b1a8bd/plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenArtifactMatcher.java) and [MavenMetadataMatcher](https://github.com/sonatype/nexus-oss/blob/556e5dc4b1adebcde278e14f30f187a023b1a8bd/plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenMetadataMatcher.java), but clearly increases confusion in MavenPath itself. Without the `Kind` the two condition could be swapped with following equivalents:

`mavenPath.getCoordinates() != null // is artifact` and `mavenPath.main().getFileName().equals("maven-metadata.xml")`

So maybe remove the Kind altogether?

cstamas: Actually, Kind is wrong and am removing it: it carries Maven2 specific information (the metadata), while MavenPath should be layout agnostic, and should work with all maven versions we are about to support.

mrprescott:
+1

cstamas:
http://bamboo.s/browse/NX3-OSSF76-1 :+1:

cstamas:
Merged as f74e69e1c6288b5c0a02067ce14d17e634cb37f5

PR #1092, opened 2015-03-09
Replaces … with &hellip; to avoid rendering problems on Windows

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8182

Related PR: https://github.com/sonatype/nexus-pro/pull/499

Extends the fix in https://github.com/sonatype/nexus-oss/pull/1085 to the remaining occurrences of `…` in the source code.


kellyrob99:
+1

PR #1093, opened 2015-03-09
[NEXUS-8115] Uses a modal instead of drilldown to mark the log

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8115

Related: https://github.com/sonatype/nexus-pro/pull/500

This PR creates a second pattern for “adding” objects, and implements it for the “Mark Log” feature.

The old pattern is *drilldown*, which is best when the create process is in any way complex. To simplify, we only animate any unneeded UI from view, so you can focus on the create form.

http://take.ms/wxyZT

The new pattern is a *modal dialog*, to be used when the create process is simple (one or two fields). Because the form is simple, we have plenty of screen space, and the user can quickly fill it and save, without enduring the heavy drilldown interaction.

http://take.ms/AzyGo


adreghiciu:
+1

PR #1100, opened 2015-03-11
Blobstore recipe

kellyrob99:
Abstract the storage implementation to a Facet, allowing for both different impls for backend storage and set the stage for any additional Facets that we might need related to BlobStore.

Fair bit of duplication here from the Repository impl, but if we're happy with this I will look into refactoring some shared pieces/logic out.


jdillon:
-1 blobstore apis does not need facets, this is added complexity w/o good reason.  having more _facets_ also sets the stage for confusion about what facet you are talking about in what framework, etc.

kellyrob99:
Closing after talking to @jdillon

PR #1102, opened 2015-03-11
[NEXUS-8204] Removes second 'x' from input boxes in IE10/11

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8204

To fix this, I had to remove the X-UA-Compatible meta tag, which was forcing IE to render in IE9 compatibility mode.

https://connect.microsoft.com/IE/feedback/details/783743/disable-ie10-clear-field-button-when-rendering-in-compatibility-mode

So far, I can’t see any negative consequences of removing that tag, but I’d definitely like another set of eyes, particularly from @jdillon.


plugins/rapture/nexus-rapture-plugin/src/main/resources/org/sonatype/nexus/rapture/internal/index.vm
```diff
@@ -20,9 +20,6 @@
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
   <meta name="description" content="Sonatype Nexus"/>
 
-  ## force browser to behave, IE=edge seems to have issues on IE 11 setting stylesheets (maybe other shit too)
-  <meta http-equiv="X-UA-Compatible" content="IE=9,chrome=1"/>
```
jdillon: Extjs5 docs suggest:

    <meta http-equiv="X-UA-Compatible" content="IE=edge">

Why don't we use that vs. removing it?

djsauble: There’s certainly no harm in being explicit here. I’ll add it back.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Icon.js
```diff
@@ -100,8 +100,6 @@ Ext.define('NX.controller.Icon', {
     });
 
     // create the style sheet
-    // NOTE: This has issues on IE 11, forced compat via X-UA-Compatible meta tag
-    // NOTE: ... but may be better off rendering this server side?
     me.stylesheet = Ext.util.CSS.createStyleSheet(styles.join(' '), 'nx-icons');
```
peterlynch: I wonder if the mentioned problems were related to how when IE sees the meta tag, it then has to start re-parsing the document again.

In case this re-parsing causes some sort of issue, there apparently is an option to return an HTTP header instead.

From [here](https://www.modern.ie/en-us/performance/how-to-use-x-ua-compatible):
> The best practice is an X-UA-Compatible HTTP Header. Adding the directive to the response header tells Internet Explorer what engine to use before parsing content begins.

jdillon:
+1

djsauble:
Manual merge. https://github.com/sonatype/nexus-oss/commit/8b7e30195dcae7976a2c5bdbdce2d8eb7fa33c60

PR #1106, opened 2015-03-12
[NEXUS-8251] Remove custom native elastic search filtering

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8251
Using custom native ES filter does not work in a clustered environment nor to access current subject via ThreadLocal due to queries / filters being executed via a thread pool.


jdillon:
+1

PR #1110, opened 2015-03-13
Add possibility to install multiple hooks to DB

cstamas:


cstamas:
Explanation to change in fa7964e4cfbdbcad2c1871ef0f6e4a562e6bedb9
Turns out that the original PR (and even master in very same way) breaks repository removal: on repository removal facets are shut down in reverse order, but while SearchFacet is shut down, the Search hook is still in place, and on StorageFacetImpl.deleteAll the hook explodes (as it tries to remove deleted vertexes from SearchFacet, while SearchFacet is already STOPPED). Clearly, addition of hooks should be somehow tied to "interested party" (facet, in this case) lifecycle, while the "global" lastUpdated hook should be always there.

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/StorageFacetImpl.java
```diff
@@ -223,65 +222,57 @@ public StorageTx openTx() {
...
-    catch (MissingFacetException e) {
-      // no search facet, no indexing
-      return new GraphTx(db);
+    final ODatabaseDocumentTx db = databaseInstanceProvider.get().acquire();
+    final HookedGraphTx graphTx = new HookedGraphTx(db);
+    for (ORecordHookFactory recordHookFactory : recordHookFactories) {
+      try {
+        graphTx.registerHook(checkNotNull(recordHookFactory.createHook(graphTx)));
+      }
+      catch (MissingFacetException e) {
```
adreghiciu: Maybe catch Exception?

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/search/SearchFacetRecordHookFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import com.orientechnologies.orient.core.hook.ORecordHook;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Factory for {@link SearchFacetRecordHook} hook.
+ *
+ * @since 3.0
+ */
+public class SearchFacetRecordHookFactory
```
adreghiciu: Why not use an inner class instead?

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/LastUpdatedRecordHookFactory.java
```diff
@@ -0,0 +1,32 @@
...
+import org.sonatype.nexus.repository.Repository;
+
+import com.orientechnologies.orient.core.hook.ORecordHook;
+
+/**
+ * Factory for {@link LastUpdatedRecordHook} hook.
+ *
+ * @since 3.0
+ */
+public class LastUpdatedRecordHookFactory
```
adreghiciu: Why not use an inner class instead?

cwilper:
Please see https://github.com/sonatype/nexus-oss/pull/1112 which fixes the repository removal issue and adds a last_updated hook, while keeping the hooks as inner classes.

cstamas:
As per phone, closing this PR in favor of Chris' PR

PR #1113, opened 2015-03-13
NEXUS-8267 - NuGet feed data contains unexpected attribute key

kellyrob99:
- trivial addition of 'alias' for DOWNLOADCOUNT property

https://issues.sonatype.org/browse/NEXUS-8267


djsauble:
What different purposes do `DOWNLOAD_COUNT` and `DOWNLOAD_COUNT_ALIAS` serve?

kellyrob99:
@djsauble in some cases the underlying data says its property name for the orderBy is DOWNLOAD_COUNT and our expectation was that it is called DOWNLOADCOUNT. This only happens for some records

djsauble:
+1

cwilper:
+1

PR #1116, opened 2015-03-17
refactor blobstore creation by provider

kellyrob99:
Allow for additional future BlobStore types

### DIFF
--- a/components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
+++ b/components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
@@ -60,6 +60,10 @@ public class FileBlobStore
   public static final String TYPE = "File";
 
   public static final String BLOB_CONTENT_SUFFIX = ".blob";
+  
+  private static final String CONFIG_KEY = "file";
+  
+  private static final String PATH_KEY = "path";
 
   private Path root;
 
@@ -294,7 +298,7 @@ public class FileBlobStore
   @Override
   public void init(final BlobStoreConfiguration configuration) throws IOException {
     this.blobStoreConfiguration = configuration;
-    Path blobDir = Paths.get(getPath(configuration.getAttributes()));
+    Path blobDir = Paths.get(String.valueOf(configuration.attributes(CONFIG_KEY).require(PATH_KEY)));
     Path content = blobDir.resolve("content");
     File metadataFile = blobDir.resolve("metadata").toFile();
     DirSupport.mkdir(content);
@@ -321,6 +325,14 @@ public class FileBlobStore
     attributes.put("path", path);
     map.put("file", attributes);
     return map;
+  }
+
+  public static BlobStoreConfiguration configure(final String name, final String path) {
+    BlobStoreConfiguration configuration = new BlobStoreConfiguration();
+    configuration.setName(name);
+    configuration.setType(FileBlobStore.TYPE);
+    configuration.attributes(CONFIG_KEY).set(PATH_KEY, path);
+    return configuration;
   }
 
   class FileBlob
--- a/components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
+++ b/components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
@@ -12,14 +12,18 @@
  */
 package org.sonatype.nexus.blobstore.file;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.FileStore;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.HashMap;
 import java.util.Map;
 
 import javax.annotation.Nullable;
+import javax.inject.Named;
 
 import org.sonatype.nexus.blobstore.api.Blob;
 import org.sonatype.nexus.blobstore.api.BlobId;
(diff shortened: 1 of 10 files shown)


kellyrob99:
+1 by CI http://bamboo.s/browse/NX3-OSSF106-1

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
```diff
@@ -44,35 +51,44 @@
...
     this.fileOperations = checkNotNull(fileOperations);
-    this.metadataStore = checkNotNull(metadataStore);
-    this.blobStoreConfiguration = checkNotNull(blobStoreConfiguration);
+  }
+
+  @VisibleForTesting
+  public FileBlobStore(final Path root, final LocationStrategy locationStrategy,
+                              final FileOperations fileOperations, final BlobMetadataStore metadataStore, 
+                              final BlobStoreConfiguration configuration)
+  {
```
jdillon: please line up args I find this very hard to read at a glance

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
```diff
@@ -282,6 +310,18 @@ private void checkExists(final Path path, final BlobId blobId) throws IOExceptio
     }
   }
 
+  private String getPath(final Map<String, Map<String, Object>> attributes) {
+    return (String) attributes.get("file").get("path");
+  }
```
jdillon: This reminds me we should extract *AttributesMap to common so we can re-use that.

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
```diff
@@ -271,7 +287,19 @@ public void compact() {
...
+
+  @Override
+  public void init(final BlobStoreConfiguration configuration) {
+    this.blobStoreConfiguration = configuration;
+    Path blobDir = Paths.get(getPath(configuration.getAttributes()));
+    Path content = blobDir.resolve("content");
+    File metadataFile = blobDir.resolve("metadata").toFile();
+    content.toFile().mkdirs();
+    metadataFile.mkdirs();
```
jdillon: should probably use DirSupport helpers here to make directories

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
```diff
@@ -44,35 +51,44 @@
...
-    this.blobStoreConfiguration = checkNotNull(blobStoreConfiguration);
+  }
+
+  @VisibleForTesting
+  public FileBlobStore(final Path root, final LocationStrategy locationStrategy,
+                              final FileOperations fileOperations, final BlobMetadataStore metadataStore, 
+                              final BlobStoreConfiguration configuration)
+  {
+    this(locationStrategy, fileOperations);
+    this.root = root;
```
mrprescott: checkNotNull?

components/nexus-core/src/main/java/org/sonatype/nexus/internal/blobstore/BlobStoreManagerImpl.java
```diff
@@ -245,22 +220,18 @@ private void untrack(final String name) {
...
    */
   private void validate(final BlobStoreConfiguration configuration) {
-    final String path = getPath(configuration.getAttributes());
-    boolean duplicatePath = Iterables.any(store.list(), new Predicate<BlobStoreConfiguration>()
-    {
-      @Override
-      public boolean apply(final BlobStoreConfiguration input) {
-        return path.equals(getPath(input.getAttributes()));
+    for (BlobStoreConfiguration blobStoreConfiguration : store.list()) {
+      if (Maps.difference(blobStoreConfiguration.getAttributes(), configuration.getAttributes()).areEqual()) {
```
mrprescott: This looks suspicious to me - I think this test works now, but it seems coincidental to the fact that FileBlobStore only has one attribute, and it must have a unique value. This test seems to imply that it's okay to have two file blob stores writing to the same location, so long as they have some other setting that differs.

kellyrob99: True, works now because there are no additional attributes

mrprescott: Perhaps: if(!blobStoreConfiguration.isCompatible( configuration) ) ?

kellyrob99: This implies that the configuration knows how to interpret all possible attribute formats, which is not the case.

kellyrob99: Once we have other types, will probably have to inject something like Map<String, Validator> corresponding to the blobStorePrototypes, signature like:
```
void validate(BlobStoreConfiguration configuration, List<BlobStoreConfiguration> existingConfigurations)
```

jdillon: This is added complexity w/o good reason, I would drop it for now.

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
```diff
@@ -271,7 +288,19 @@ public void compact() {
 
   @Override
   public BlobStoreConfiguration getBlobStoreConfiguration() {
-    return blobStoreConfiguration;
+    return this.blobStoreConfiguration;
+  }
+
+  @Override
+  public void init(final BlobStoreConfiguration configuration) throws IOException {
+    this.blobStoreConfiguration = configuration;
```
jdillon: probably checkNotNull here for sanity?

jdillon:
+ overall *vastly* simpler than previous attempt thank you!

mrprescott:
+1

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
jdillon: this is merged to nexus-common now so you can use AttributesMap directly now

kellyrob99:
+1 by CI

PR #1115, opened 2015-03-16
[NEXUS-8278] Component/Asset wrappers for OrientVertex

cwilper:
https://issues.sonatype.org/browse/NEXUS-8278

PR #1117, opened 2015-03-18
[NEXUS-8015] Add support for write policy

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8015


cstamas:
+1

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/IllegalOperationException.java
```diff
@@ -0,0 +1,26 @@
...
+ * which accompanies this distribution and is available at http://www.eclipse.org/legal/epl-v10.html.
+ *
+ * Sonatype Nexus (TM) Professional Version is available from Sonatype, Inc. "Sonatype" and "Sonatype Nexus" are trademarks
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+package org.sonatype.nexus.repository;
+
+/**
+ * Thrown when an illegal operation has been encounter.
```
kellyrob99: minor spelling: encountered

kellyrob99:
+1 but could use some basic unit testing

adreghiciu:
Will wait till sonatype/nexus-oss/pull/1115 is merged to adapt to those changes and then make the UTs

cwilper:
#1115 is merged. Looking at this, I'm curious why it only covers blobs. Is it the intent that asset and components can still be added/modified/deleted?  (except this would prevent assets that have blobs from having those blobRefs changed or having the asset deleted)

jdillon:
@cwilper good point.  I think the intent of the policy is to allow components/assets to be created/updated, or just created, or not created at all.

cwilper:
ok. So technically this code as-is would allow attributes of components/assets to be modified regardless of the policy setting. One could get a reference to a Component or Asset, modify properties at will, then commit. If it were just "read only" or "read/write", it would be easy to enforce across the board at this level, as mentioned, you could just prevent commits. But the "append only" nature of the "ONCE" policy makes it a bit more nuanced. I don't have a great solution to that, unless Components and Assets are constructed with knowledge of the policy in play, and can then prevent set operations if isNew() is false...? dunno.

jdillon:
@cwilper attributes should still be mutable.  The point of this is to have a policy that can allow content of a repository to be deployed/uploaded/created one time, many times, or not at all.  Ie. if you have a ONCE policy on a release repo and you upload foo:bar:1:jar, you should not be allowed to upload that same component/asset again.  On a repo with policy of MANY, you can upload foo:bar:1:jar over and over its just fine.  And with DENY you can't upload anything.

Internally we many need to mutate the attributes for book keeping.  So this is really more about a policy from the repo clients perspective.  Probably also some concern about if components _move_ from repo to repo, so that if you moved foo:bar:1:jar from repo-a to repo-b and repo-b already had this component with a policy of ONCE or DENY then this operation should fail, because it already exists and can't be changed.

cwilper:
@jdillon thanks for clarifying the policy in detail. I think this level of enforcement is fine then.

cwilper:
+1 after UTs

adreghiciu:
UT's added + policy is not checked when bucket is deleted (repo is deleted)

adreghiciu:
Manually merged https://github.com/sonatype/nexus-oss/commit/c04e68d22851d8ae6d62362104ee90165c09856a

PR #1122, opened 2015-03-19
NEXUS-8281: HEAD support in sender and some crucial Handlers

cstamas:
Issue
https://issues.sonatype.org/browse/NEXUS-8281

### DIFF
--- a/plugins/basic/nexus-repository-httpbridge/src/main/java/org/sonatype/nexus/repository/httpbridge/DefaultHttpResponseSender.java
+++ b/plugins/basic/nexus-repository-httpbridge/src/main/java/org/sonatype/nexus/repository/httpbridge/DefaultHttpResponseSender.java
@@ -17,13 +17,17 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Map;
 
+import javax.annotation.Nullable;
 import javax.inject.Named;
 import javax.inject.Singleton;
 import javax.servlet.ServletException;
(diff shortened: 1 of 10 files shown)


plugins/basic/nexus-repository-httpbridge/src/main/java/org/sonatype/nexus/repository/httpbridge/DefaultHttpResponseSender.java
```diff
@@ -48,7 +51,7 @@
     implements HttpResponseSender
 {
   @Override
-  public void send(final Response response, final HttpServletResponse httpResponse)
+  public void send(final @Nullable Request request, final Response response, final HttpServletResponse httpResponse)
```
kellyrob99: Minor, but technically all we need is the 'action' here, not the entire request

jdillon: This impl only needs that, unsure if others may or may not which is why I suggested just pass in the request and leave it up to the impl to sort out what it needs.

FTR I also think we could make this non-nullable in the future, but for now I think this is sufficient.

jdillon:
+1

kellyrob99:
+1

adreghiciu:
+1

cstamas:
http://bamboo.s/browse/NX3-OSSF112-1 :+1:

PR #1126, opened 2015-03-20
Decorated payload experiment

cstamas:
This is just an experiment, a "mind game" with decorating payload. While we do have `Response` that is able to achieve pretty much the same as this change, problems with it are:
* `Response` would drag HTTP semantics into facets, it is very HTTP-like object, similarly as `Request` that has processing mostly encapsulated in `Handler`s. You should use HTTP headers for things that might not be headers at all, and you'd need to mind the HTTP formatting too (of dates, etc).
* `Response`, as seen on current format implementations, rarely is returned directly by facets, it is usually crafted by `Handler`s. And in many cases, you don't want your own format specific handler (proxy, group?).
* Format specific payloads are unwanted.

This experiment shows one possibility, to "decorate" payloads, make them spicier with a bit of extra information, that might be passed from things like `XXXStorageHandler` to `XXXProxyFacet` and other way around. I believe while not obvious in case of hosted repositories, this might be useful in case of proxy and group repository implementations...

Examples of uses of this change:
* Maven Proxy: IF/WHEN we decide to make Maven Proxy fetch preemptively sha1/md5 files, those could be just another decorator sent from `MavenProxyFacet` to `MavenFacet` for storing, where latter could apply some extra logic/verify and store them (instead as now, we'd need to actually fetch the sha1/md5 hashes two times: once to validate, once to cache).
* Maven group: IF we decide to cache merged metadata, we could easily use event listeners on group cache to invalidate cached entries by extra data provided via each member's "spice", like SHA1 hash of each merged metadata file (content change detection).
* Non-default HTTP response sender: if format needs some extra bits to send to it's own specific HTTP response sender but `Response` can't be used (like the case was with Maven specific headers from proxy/group), it can be now easily done, without dragging all the HTTP semantics/header formatting stuff into place where it should not be (non HTTP specific facet for example). Also, with these changes default response sender should be easier for reuse.


plugins/basic/nexus-repository-httpbridge/src/main/java/org/sonatype/nexus/repository/httpbridge/DefaultHttpResponseSender.java
```diff
@@ -50,62 +61,180 @@
...
+    if (payload instanceof DecoratedPayload) {
+      final DecoratedPayload decoratedPayload = (DecoratedPayload) payload;
+      handleDecoratedPayloadHeaders(decoratedPayload, httpResponse);
+    }
+  }
+
+  /**
+   * Set HTTP response headers based on {@link DecoratedPayload}.
+   */
+  protected void handleDecoratedPayloadHeaders(final DecoratedPayload decoratedPayload,
```
mrprescott: This is a slightly misleading method name, since it deals with not just PayloadHeaders but a lot of other decorations.

plugins/basic/nexus-repository-httpbridge/src/main/java/org/sonatype/nexus/repository/httpbridge/DefaultHttpResponseSender.java
```diff
@@ -50,62 +61,180 @@
...
+    final PayloadTimes payloadTimes = decoratedPayload.decorator(PayloadTimes.class);
+    if (payloadTimes != null) {
+      addHeaderIfNotPresent(
+          httpResponse,
+          HttpHeaders.LAST_MODIFIED,
+          payloadTimes.getLastModified()
+      );
+    }
+    // ETag using hashes
+    final PayloadHashes payloadHashes = decoratedPayload.decorator(PayloadHashes.class);
```
mrprescott: This big method seems to make it really hard for format-specific senders to do custom thing - for example, maven might set the ETag differently.

jdillon:
@cstamas why wouldn't you just use the response.attributes to encode more information?

plugins/basic/nexus-repository-httpbridge/src/main/java/org/sonatype/nexus/repository/httpbridge/DefaultHttpResponseSender.java
```diff
@@ -50,62 +61,180 @@
...
+   * Set HTTP response headers based on {@link Response}.
+   */
+  protected void handleResponseHeaders(final Response response,
+                                       final HttpServletResponse httpResponse)
+  {
+    // add response headers
+    for (Map.Entry<String, String> header : response.getHeaders()) {
+      httpResponse.addHeader(header.getKey(), header.getValue());
+    }
+  }
```
jdillon: curious why are you making protected methods for a class that is never sub-classed?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacet.java
```diff
@@ -40,7 +40,7 @@
   ChecksumPolicy getChecksumPolicy();
 
   @Nullable
-  BlobPayload get(MavenPath path) throws IOException;
+  DecoratedPayload get(MavenPath path) throws IOException;
```
jdillon: probably don't want to expose anything but a Payload in api

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/payloads/PayloadHeaders.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Payload headers.
+ *
+ * @since 3.0
+ */
+@Immutable
+public class PayloadHeaders
+{
+  private Headers headers;
```
jdillon: Headers is *not* immutable, so your use of @Immutable is erroneous.

jdillon: FTR that does not mean add complexity to make adapters for Headers that are immutable, just means we don't need this complexity of an _immutable_ assurance here.

cstamas: `PayloadHeaders` is wrong, should not exists (why would we add Headers to it, exact ones that Response has too), @mrprescott also pointed over this over HC, neglect this class.

cstamas:
@jdillon As I wrote in PR description, currently no repo format assembles response in facet, only handlers. This would mean we'd need format specific handlers across many formats only to achieve the same thing, that is currently done by same (core) handlers. Also, we strieve to reuse handlers for proxy and group. 

And these "decorations" (like times, hashes) are not HTTP specific at all, while Responses are made in handlers via HttpResponses are... actually we agreed to have HTTP specific handlers, but if a facet emits Response, doesn't it make facet HTTP specific too then?

cstamas:
Closing as we want to keep sender "slim" as much possible, while this one drifts in opposite direction.

PR #1128, opened 2015-03-20
NEXUS-8178: Maven CMA group

cstamas:


plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -188,9 +208,27 @@ public void put(final MavenPath path, final Payload payload)
     try (StorageTx tx = getStorage().openTx()) {
       if (path.getCoordinates() != null) {
         putArtifact(path, payload, tx);
+        if (payload instanceof MavenPayload) {
```
kellyrob99: Couple of places we cast but do nothing if we are passed the wrong type, and I'm thinking that if we get here and have something other than a MavenPayload, it should be an error condition.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MetadataMerger.java
```diff
@@ -0,0 +1,352 @@
...
+    if (result.getVersioning() != null && !result.getVersioning().getVersions().isEmpty()) {
+      Collections.sort(result.getVersioning().getVersions(), versionStringComparator);
+      result.getVersioning().setLatest(getLatestVersion(result.getVersioning().getVersions()));
+      result.getVersioning().setRelease(getReleaseVersion(result.getVersioning().getVersions()));
+    }
+    if (!result.getPlugins().isEmpty()) {
+      Collections.sort(result.getPlugins(), pluginComparator);
+    }
+
+    // model version, just set it to latest, we don't care about it now
```
kellyrob99: Is there a point that we will care about it?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MetadataMerger.java
```diff
@@ -0,0 +1,352 @@
...
+        lv.getSnapshotVersions().add(newSnapshotVersion);
+      }
+    }
+  }
+
+  /**
+   * Returns last element of sorted string version list, the "latest".
+   */
+  private String getLatestVersion(final List<String> sortedVersions) {
+    return sortedVersions.get(sortedVersions.size() - 1);
```
kellyrob99: index problems for an empty list

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MetadataMerger.java
```diff
@@ -0,0 +1,352 @@
...
+      return "";
+    }
+    return str.trim();
+  }
+
+  /**
+   * Filters a list for {@code null} (and "null" string) elements, removes them.
+   */
+  private <T> void nullElementFilter(final List<T> list) {
+    for (Iterator<T> i = list.iterator(); i.hasNext(); ) {
```
kellyrob99: minor but you could also use Guava to strip nulls, although it does return a view instead of modifying the original collection:
```
return Collections2.filter(list, Predicates.notNull())
```

plugins/basic/nexus-repository-maven/src/test/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MetadataMergerTest.java
```diff
@@ -0,0 +1,183 @@
...
+    sources.setClassifier("sources");
+    sources.setVersion(versionPrefix + "-" + timestamp + "-" + buildNumber);
+    sources.setUpdated(timestamp);
+    mv.getSnapshotVersions().add(sources);
+
+    return m;
+  }
+
+  @Test
+  public void smokePlugin() throws Exception {
```
kellyrob99: What are YOU smoking?
But seriously, could use some better method names ;)

cstamas:
Replaced by https://github.com/sonatype/nexus-oss/pull/1143 (as GH PRs cannot be "retargeted", this one was agains proxy branch that is now in master)

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MetadataMerger.java
cstamas: Only if we decide to support Maven versions older than 2.0.10

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MetadataMerger.java
cstamas: It is protected, but yes, is not apparent, see where this method is invoked.

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MetadataMerger.java
cstamas: And I want to remove `"null"`s too, Strings having value of "null" as some older metadata may contain them, at least Maven code handling metadata does the same.

plugins/basic/nexus-repository-maven/src/test/java/org/sonatype/nexus/repository/maven/internal/maven2/Maven2MetadataMergerTest.java
cstamas: ack

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
cstamas: This part goes away if we get https://github.com/sonatype/nexus-oss/pull/1144 in place, I dislike it too very much

PR #1132, opened 2015-03-23
[NEXUS-8235] Adds try/catch block when loading drilldown stores

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8235

Before: http://take.ms/QZp4C

After: http://take.ms/Qexbi


kellyrob99:
+ 0.5 seems reasonable, but I haven't yet had a chance to look and see if ExtJS has any facility for configuring ExtDirect error handling in a global fashion.

djsauble:
@kellyrob99 In theory, you should be able to listen for "exception" events on Ext.direct.Manager, but the exceptions caught here don’t seem to trigger that event. And even if they did, I’m not sure it would prevent ExtJS from entering an unrecoverable state.

http://docs.sencha.com/extjs/4.2.3/#!/api/Ext.direct.Manager

jdillon:
+1

PR #1134, opened 2015-03-24
NEXUS-8236 - Store NuGet package feed metadata at the asset level

kellyrob99:
- store only top level attributes on the component, everything else at the Asset level

https://issues.sonatype.org/browse/NEXUS-8236


kellyrob99:
:+1: by CI http://bamboo.s/browse/NX3-OSSF124-1

cstamas:
+1

PR #1137, opened 2015-03-25
Pull out blob creation from TX

cstamas:
As it causes long tx windows, as content is getting streamed
down from upstream/downstream. This change allows out of tx
blob creation, leaving potential cleanup to blob GC, but leaves
room for manual cleanup.

BlobHandle is also simpler resource to reuse in case of re-attempting
tx if optmistic tx fails (no need for actual content, that is not reusable in case of HttpPayload).

This change minimizes tx windows, but does not fix the bucket
problem still.


mrprescott:
Automatic cleanup is really nice, so I'm reluctant to lose it.
I think we need to understand whether the long-running tx is actually the problem; Chris made a good point, which was that the tx duration may be irrelevant until we start touching Orient resources, because of Orient's reliance on MVCC.
It's probably worth writing a test that exposes the problem so we can measure the effect of various proposed changes.

cstamas:
I agree about automatic cleanup loss. Still, that solution does rely on fact that Orient TX is "lazy" (starts when "touching" the orient resources) which is implementation detail of orient (is it actually documented? are we sure it will not change in 2.1.x or 2.2.x?), and will result in _unwritten rule_ of something like "code to do streaming first, and THEN tamper with orient resources". And I dislike unwritten rules, as they tend to be forgotten, are not enforced.

PR #1140, opened 2015-03-26
[NEXUS-7317] Warn user when there are unsaved changes

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-7317

Updated demo: http://take.ms/c0p0F


jdillon:
cool!  should we use button styles to indicate to help distinguish the options?  unsure what to do but perhaps danger for the option that is potentially going to clobber non-saved data or something?

djsauble:
@jdillon Good point. Looks like they use a red button for the danger option.

screen shot 2015-03-25 at 7 37 03 pm

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/SettingsForm.js
```diff
@@ -95,6 +103,22 @@ Ext.define('NX.controller.SettingsForm', {
     }
   },
 
+  /*
+   * @private
+   * Reset the dirty state of the form
+   */
+  resetDirtyState: function (form) {
```
kellyrob99: This looks like it's doing an end run around the available API, and a google search suggests we may be able to do this more easily (if it's recent enough for our version of ExtJS of course :) )
http://snehalvetal.blogspot.ca/2012/08/extjs-clearing-dirty-flags-for-form.html

djsauble:
Progress, but this is becoming a bit complicated (i.e. a ball of worms).

http://take.ms/cuV8N

Components I monitor for navigation events:

+ Feature menu
+ Mode buttons
+ Browser tab (closing it)
+ Browser refresh button
+ Application refresh button
+ Modals (closing them)
+ Drilldown breadcrumb

Components I do *not* monitor:

+ Browser back button
+ Backspace (which behaves the same as clicking the browser back button)

I can easily disable backspace, and probably will unless convinced otherwise. The browser back button is difficult (or impossible) to disable, and I’m hesitant to change that behavior anyway. Thoughts?

Other bugs:

+ The “discard” button needs danger styling
+ Some forms show as dirty when they aren’t (@kellyrob99’s suggestion may fix this)

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/SettingsForm.js
djsauble: Thanks @kellyrob99. I’ll look into this. Hopefully it’ll fix a few forms that still show as dirty even though they aren’t (Repository details, Route details).

mhorvat-sonatype:
We are going to phase out the red style for 'dangerous' actions, instead making them tertiary buttons. The rationale for that is that the red is very prominent and attracts the eye, making it more likely that a user who isn't paying attention will click on it. Better to make the user read the button. however in this case it's not straightforward if 'continue' or 'go back' should be the primary button, or if both should be the same. Let's discuss.

Re Back button: have you considered https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload

You can't style the dialog or conditionally check for dirty fields, but you can warn the user they might loose changes if they have unsaved input in the form.

djsauble:
@mhorvat-sonatype `onbeforeunload` doesn’t work for hash navigation.

If I start at google.com, navigate straight to a settings page in Nexus, and edit a field, clicking the back button gives me a warning.

http://take.ms/B6fXw

However, if I’m just navigating within the app, clicking back doesn’t trigger the event.

http://take.ms/kC3Av

kellyrob99:
@whyjustin can correct me if needed, but CLM utilizes an event based model to handle this, which I believe also has dependencies on the angularjs router.
Setup: https://github.com/sonatype/insight-brain/blob/master/insight-brain-service/src/main/wro4j/assets/MainModule.js#L156-156
Consume events: https://github.com/sonatype/insight-brain/blob/master/insight-brain-service/src/main/wro4j/applications/ApplicationController.js#L335-335

kellyrob99:
@djsauble Looks like it might be missing some features in 4.2.3 that have been addressed in 5.x versions, but Ext.util.History might be the correct place for dealing with cross-cutting events.
http://docs.sencha.com/extjs/4.2.3/#!/api/Ext.util.History

djsauble:
@kellyrob99 @jdillon @mhorvat-sonatype 

This is ready for a final round of review. I’ve updated [https://issues.sonatype.org/browse/NEXUS-7317](NEXUS-7317) to describe the actual behavior of this feature.

Demo: http://take.ms/c0p0F

jdillon:
+1 a few minor questions about me vs. this

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/SettingsForm.js
djsauble: Fixed.

whyjustin:
> CLM utilizes an event based model to handle this, which I believe also has dependencies on the angularjs router

Correct, we utilize our ui routing framework (https://github.com/angular-ui/ui-router) to capture any change to the browser location.

djsauble:
Thanks @whyjustin.

Yeah, I think it’s fine to overlook the back button for this round. ExtJS 5 won’t help us preempt changes to browser history either. Should we open a ticket with Sencha?

PR #1144, opened 2015-03-26
Content gets attributes

cstamas:
Makes `MavenContent` gone, and opens possibility for proxy facet to give extra information (like would be origin URL, TBD) of the content it got and needs to be cached.


plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenHashCodes.java
```diff
@@ -19,44 +19,35 @@
...
 
 /**
- * Maven content.
+ * Maven hash codes map, usable as attribute.
  *
  * @since 3.0
  */
-public class MavenContent
-    extends Content
+public class MavenHashCodes
```
adreghiciu: What is the advantage of having this class vs. directly storing the hashes map?

jdillon: Agree this seems like a pointless wrapper.  Also hash-codes are not really specific to maven.  So perhaps consider an attribute that attaches these generally?

cstamas: Ack, I agree. @adreghiciu removing wrapper and moving them to more general place.

cstamas:
Superseded by https://github.com/sonatype/nexus-oss/pull/1146

PR #1152, opened 2015-03-31
Exploring removal of magic checksum handling for maven repositories

jdillon:
https://issues.sonatype.org/browse/NEXUS-8357
http://bamboo.s/browse/NX-OSSF692


jdillon:
dropping

PR #1159, opened 2015-04-02
Fixes bugs with the “unsaved changes” modal

djsauble:
Issue: [NEXUS-7317](https://issues.sonatype.org/browse/NEXUS-7317)

+ Prevents MultiSelect fields within our custom ItemSelector from triggering the dirty state

Before: http://take.ms/lGjJZ
After: http://take.ms/TdaTV

+ Triggers 'unsaved changes' modal for relevant actions (enable/disable capability, run/stop task, etc…)

Before: http://take.ms/94F0O
After: http://take.ms/S5ogd

+ Resets unsaved changes flag once discard handler has run. Otherwise, modal will not reappear, even if the form becomes dirty again.

Before: http://take.ms/8yJjr
After: http://take.ms/1k5oG

+ Resets forms in a drilldown which aren’t visible. Otherwise, they trigger the dirty state when you nav away, even though you’ve already chosen to discard the changes.

Before: http://take.ms/wKDOp
After: http://take.ms/bETwM


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/ext/form/field/ItemSelector.js
```diff
@@ -110,6 +110,7 @@ Ext.define('NX.ext.form.field.ItemSelector', {
       // so override these methods to prevent them from including
       // any of their values
       submitValue: false,
+      isDirty: Ext.emptyFn,
```
kellyrob99: this seems strange, explain please?

djsauble: ItemSelector consists of two MultiSelect controls. These controls are used to set the value of ItemSelector, but aren’t treated like normal form fields themselves. I override the isDirty function here, since we don’t care if they’re dirty.

kellyrob99:
+1

PR #1160, opened 2015-04-03
[NEXUS-8342] initial bits to support exception mapping

jdillon:
https://issues.sonatype.org/browse/NEXUS-8342


adreghiciu:
Decided to not implement it. see issue for details.

PR #1161, opened 2015-04-03
[NEXUS-8304,NEXUS-8231] Add bean mapping for configuration and validation support

jdillon:
https://issues.sonatype.org/browse/NEXUS-8304
https://issues.sonatype.org/browse/NEXUS-8231


jdillon:
Also added FacetSupport.facet(), since getRepository().facet() was/is used all over, can reduce a tiny bit of verbosity at the expense of a tiny helper overhead.

FTR I don't think we want to add a lot of these, NX2 did that and ended up with a lot of this sort of thing, that isn't the intent here, but clearly facet lookup is common so I think it makes sense to simplify for this specifically.

jdillon:
http://bamboo.s/browse/NX3-OSSF150-2 :white_check_mark:

jdillon:
Will probably have to sort out how violations of separate facets gets aggregated together with proper keys to apply to the UI...

adreghiciu:
+1

mrprescott:
+1

kellyrob99:
+1

cstamas:
+1

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/StorageFacetImpl.java
```diff
@@ -58,19 +60,36 @@
...
 
-  private ORID bucketId;
+  @VisibleForTesting
+  static class Config
+  {
+    @NotEmpty
+    public String blobStoreName = "default";
 
-  private WritePolicy writePolicy;
+    //@NotNull
```
kellyrob99: ?

jdillon: This is nullable atm, though probably should not be.

jdillon:
manually merged

PR #1162, opened 2015-04-03
Styling pass (implement containers and polish rough edges)

djsauble:
The primary purpose of this PR is to implement the container pattern around content. @mhorvat-sonatype 
has done a lot of work to create a pattern which makes sense for CLM, so I’ve worked with him to adopt this pattern to Nexus in a way that’s consistent and improves the Nexus UI as a whole.

screen shot 2015-04-08 at 11 02 53 am
**Container benefits**

+ Scrolling goes to the margins ([new](http://take.ms/W8wsV) vs. [old](http://take.ms/mMycG))
+ Top border makes clear what area scrolls ([new](http://take.ms/6A7kv) vs. [old](http://take.ms/hvAG1))
+ Visual hierarchy is clearer ([new](http://take.ms/Jdqwx) vs. [old](http://take.ms/faTPh))
+ Light-grey background color draws attention to the content ([new](http://take.ms/YsgPd) vs. [old](http://take.ms/ls4lF))
+ Feature buttons don’t float anymore ([new](http://take.ms/XOUzm) vs. [old](http://take.ms/dtqfZ))
+ Loading mask is anchored ([new](http://take.ms/Pi2Tr) vs. [old](http://take.ms/kBvck))

**Change log**

+ Wraps headers, forms, feature groups, info panels, and other blocks of content (e.g. metrics, system information) in containers
+ Border added to the top of every feature (either directly below the breadcrumb, or below the top toolbar)
+ Background of content area changed to `#F4F4F4`
+ Darkens menu to better contrast with the darker content background
+ Renames `nx-drilldown-action` to `nx-actions`, since the toolbar is used on non-drilldown views
+ Abstracts SSL certificate details into separate views, per the SettingsPanel/SettingsForm pattern
+ Converts `tbar` config shorthand to `dockedItems`, so the toolbar can be styled
+ Sets `dockedItems` to null in views that already have a toolbar defining the space
+ SettingsForm now initializes `settingsForm` instead of `items` directly, so SettingsPanel can insert padding (`nx-inset`) dynamically.
+ Removes NX.ext.panel.SubsectionPanel view, since we can just use an Ext.Panel instead
+ Adds field label to the realm settings form
+ Fixes button alignment in modals
+ Fixes glyph cropping in file upload buttons (by reducing the glyph font size)
+ Aligns modal header text to the form and buttons inside the modal


plugins/rapture/nexus-coreui_legacy-plugin/src/main/resources/static/rapture/NX/coreui_legacy/controller/Repositories.js
```diff
@@ -112,7 +112,7 @@ Ext.define('NX.coreui_legacy.controller.Repositories', {
           beforedestroy: me.stopStatusPolling
         },
         'nx-coreui_legacy-repository-feature button[action=browse]': {
-          click: me.navigateToBrowseMode
+          runaction: me.navigateToBrowseMode
```
kellyrob99: This change seems both counter-intuitive and not consistently applied: why are we moving from a 'known' event like click to runaction and why only for certain buttons?

kellyrob99:
Two comments for the modals: titles should likely be aligned as well, and we should probably ditch the 'X' for closing in favor of the 'Cancel' button universally.
9ebac97c-da04-11e4-874a-72753584c67e

kellyrob99:
Lots of seemingly unrelated changes here, looks like we updated nx-actions, changed some layouts, and refactored logic for some dirty checking cases? In addition a lot is just formatting which could happen outside of the PR.
+0 as I'm not entirely sure which goal(s) were achieved by this - let alone how to test it and ensure regressions haven't been introduced.

plugins/rapture/nexus-coreui_legacy-plugin/src/main/resources/static/rapture/NX/coreui_legacy/controller/Repositories.js
djsauble: Because I trigger navigation from the “handler” config option, which is in the view, while the code to perform mode switching is in the controller. I use the “runaction” event to pass control from the view to the controller.

Why don’t I just override the “click” handler? When a form is dirty, I cancel navigation. I do this by returning “false” from the click event handler (which prevents the “handler” function, and–by proxy–the “truncation” event). If I used “click” for both dirty checking and navigation, this would not work.

And yeah, it’s really fugly. Any suggestions?

djsauble:
@kellyrob99 I merged https://github.com/sonatype/nexus-oss/pull/1159 into this PR, which is why those changes show up. It’s a single commit (https://github.com/sonatype/nexus-oss/commit/2692c6e1531548b2ec6c8cfbabc1984bc9bb7ac2), so I can revert it if that PR doesn’t go through.

I plan to document the other changes more thoroughly before this hits review. It’s a lot of layout/styling changes across the entire application, so testing this will be tricky.

djsauble:
Added more definition around this PR. See the updated description at the top.

kellyrob99:
+1, please squash when merging and we'll sort out how to validate the changes later

PR #1171, opened 2015-04-09
NEXUS-8435: Alternate approach to fix StorageFacet container contention

cstamas:
Changes:
* Blob-fix: Similar to PR https://github.com/sonatype/nexus-oss/pull/1137 but retains blob automatic cleanup. Makes possible to stream HTTP response without touching (and hence, starting implicitly TX) OrientDB entities. No need to load up anything from DB, and risk MVCC version becomes stale during streaming, see MavenFacetImpl.
* Bucket-contention fix: Handles assets and components as before, but defers link creation with logic to retry, happens seamlessly w/o caller noticing anything. This makes bucket contention window very small, and handled in a manner proposed by Orient folks: by retrying the Tx (but only of the link creation).
* in worst case, this code will leave components and assets "orphans" (w/o owning bucket), and those should be GCed in very same way as we handle "orphan" Blobs. Also, they are not getting in a way of existing code, as all READ ops always start with a Bucket having edge to asset or component.
* if caller gets OConcurrentModificationEx, caller it the one that should behave as appropriate for his use case. He can get it due to a) concurrent modification of asset or component on his behalf, or b) inability to create links. In both cases, the failed Tx may leave orphan assets and components, that should be GCed.

Tested with Maven repositories.


components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/BlobHandle.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.sonatype.nexus.blobstore.api.BlobRef;
+import org.sonatype.nexus.common.hash.HashAlgorithm;
+
+import com.google.common.hash.HashCode;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Blob handle.
```
mrprescott: Be good to have a bit of javadoc that explains what this is for, in particular the 'attached' attribute.

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/StorageTxImpl.java
```diff
@@ -84,6 +107,7 @@ public StorageTxImpl(final BlobTx blobTx,
     this.graphTx = checkNotNull(graphTx);
     this.bucketId = checkNotNull(bucketId);
     this.writePolicy = writePolicy;
+    this.links = Lists.newArrayList();
```
mrprescott: I suppose this implies that the links aren't traversable during the transaction they're being created.

mrprescott:
+1, pending conversation about document vs. graph api usage :P

jdillon:
-1, thought we had a solid recommendation to move to document api, unsure why we are now dancing around this.

cstamas:
Closing out as per phone.

PR #1173, opened 2015-04-09
NEXUS-8016 Group member compatibility

mrprescott:
https://issues.sonatype.org/browse/NEXUS-8016


cstamas:
+1, but this change does not really ensure that group repositories can only have compatible constituent repos nor relies on validation API :smile:

mrprescott:
Agreed, it's a reduction in scope essentially.

mrprescott:
http://bamboo.s/browse/NX3-OSSF162-1 :white_check_mark:

jdillon:
@cstamas the UI can/should ensure that you can't configure invalid repos in a group, the logic here is just to prevent manually tampered configuration from doing things it shouldn't.  We can revisit later but this is the minimal simplest change to meet the underlying needs.

jdillon:
+1 much simpler

PR #1175, opened 2015-04-10
NEXUS-8361 repo config ui

kellyrob99:
https://issues.sonatype.org/browse/NEXUS-8361


plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/services/RepositoryFormService.js
```diff
@@ -0,0 +1,268 @@
...
+Ext.define('NX.coreui.services.RepositoryFormService', {
+  singleton: true,
+  mixins: [
+    'NX.LogAware'
+  ],
+  requires: [
+    'NX.I18n'
+  ],
+
+  facets: {
```
jdillon: this isn't going to scale too well.

kellyrob99: I did some reading up on ExtJS class system this weekend and am thinking that this simple 3 method 'interface' could be handled by a mixin, separating out the 'facets' per recipe and providing a registry keyed on the recipe name - something similar to the Icon controller. Still not entirely sure how to get then to dynamically register in ExtJS, but I'm sure we'll figure something out :)
How does that sound to you @jdillon ?

PR #1178, opened 2015-04-13
[NEXUS-8408] In-app tutorials (DEMO)

djsauble:
This PR is for reference only. We will not be implementing in-app tutorials in NX 3.0. This is mainly to demonstrate how such an experience might work in the future.

Story: https://issues.sonatype.org/browse/NEXUS-8408

Most recent screencast: http://take.ms/8ljET


djsauble:
Updated screencast: http://take.ms/702kF

djsauble:
Final screencast: http://take.ms/8ljET

PR #1185, opened 2015-04-15
Allow creation of composite keys index names

adreghiciu:
* Allows setting multiple properties for an index name to be used with composite keys.
* Use "_" instead of "." after type as <type>.<name> is a special meaning for Orient (http://orientdb.com/docs/last/Indexes.html)

http://bamboo.s/browse/NX3-OSSF174


mrprescott:
+1

cstamas:
+1

kellyrob99:
+1

PR #1187, opened 2015-04-15
Avoid duplicate components/assets by using unique indexes

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8346
https://issues.sonatype.org/browse/NEXUS-8347

For components I considered unique bucket/group/name/version.
For assets I considered unique bucket/component/name.
I also made asset name mandatory, for raw/nuget I assigned the name to component name.

The index creation is different then usual since we have to instruct orient to index null values.
That is done by specifying additional metadata: ignoreNullValues = false.
The other property mergeKeys = false, is added just to avoid orient db NPE

http://bamboo.s/browse/NX3-OSSF176


cstamas:
+1

kellyrob99:
+1

adreghiciu:
CI :+1: http://bamboo.s/browse/NX3-OSSF176-1

PR #1195, opened 2015-04-16
NEXUS-8305: Simple component visitor

cstamas:
A simple component visitor. Given assumptions that component may have ten-ish assets, it is assumed that visitor implementation is able to "grab" all them using calls like `tx.browseAssets(Component)` or alike and perform processing as it needs.

Issue


components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/ComponentCursor.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    void close();
+  }
+
+  /**
+   * Opens the cursor and returns {@link Cursor}, never {@code null}.
+   */
+  @Nonnull
+  Cursor open();
+}
```
jdillon: Naming odd, a cursor with an embedded cursor?

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/Visitor.java
```diff
@@ -0,0 +1,43 @@
...
+ * A generic node visitor.  No assumptions should be made agains tx passed in, it may be same or
+ * different than passed into other methods.
+ *
+ * @since 3.0
+ */
+public class Visitor<T>
+    extends ComponentSupport
+{
+  /**
+   * Always invoked before visiting begins.
```
mrprescott: Before all visiting, or before each node?

cstamas:
Visiting/looping components as general solution on hold for now.

PR #1196, opened 2015-04-16
NEXUS-8419 - NPE on component deletion

mrprescott:
https://issues.sonatype.org/browse/NEXUS-8419
(Note, NEXUS-8423 is a dupe of NEXUS-8419, hence the incorrect branch name)


kellyrob99:
Pretty subtle change - possible to add a confirming test? Otherwise +1

components/nexus-orient/src/main/java/org/sonatype/nexus/orient/entity/EntityAdapter.java
```diff
@@ -185,8 +185,6 @@ protected void deleteEntity(final ODatabaseDocumentTx db, final T entity) {
 
     ORID rid = recordIdentity(entity);
     db.delete(rid);
-
-    entity.setEntityMetadata(null);
```
jdillon: why?  after this point the entity is toast, its md should be null'd

jdillon:
-1 please do not merge, I believe the bug is in how the entity is being used after its deleted, not in how the metadata for entities is detached upon deletion.

adreghiciu:
-1
To solve the issue we should store the entity id and change search/facet service to work with id instead of component.
Alternative would be to call search service before removing component but then we can run into problems that component is deleted from index but component is falling to delete.

PR #1199, opened 2015-04-16
NEXUS-8234 - Blob iteration

mrprescott:
https://issues.sonatype.org/browse/NEXUS-8234
This turned out way simpler than I imagined. Unsure if we need to make some sort of Visitor for it, but I figure that can wait until it's obvious.

### DIFF
--- a/components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
+++ b/components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
@@ -60,9 +60,9 @@ public class FileBlobStore
   public static final String TYPE = "File";
 
   public static final String BLOB_CONTENT_SUFFIX = ".blob";
-  
+
   private static final String CONFIG_KEY = "file";
-  
+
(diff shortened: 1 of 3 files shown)


cstamas:
+1

adreghiciu:
+1

mrprescott:
http://bamboo.s/browse/NX3-OSSF190-1 :white_check_mark:

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/FileBlobStore.java
```diff
@@ -307,6 +307,11 @@ public void init(final BlobStoreConfiguration configuration) throws IOException
     this.metadataStore = MapdbBlobMetadataStore.create(metadataFile);
   }
 
+  @Override
+  public AutoClosableIterable<BlobId> iterator() {
+    return metadataStore.findWithState(BlobState.ALIVE);
```
jdillon: why only alive?  and/or why not expose the state to iterate?

asking because one use of this is for admin/debug so it may be useful to have the ability to see all the blobs, not just those that are alive.

PR #1200, opened 2015-04-17
[NEXUS-8057] Adds patch from Sencha for double-click resize bug

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8057

Sencha sent me a patch for this bug (!). I’m filing a ticket for another issue Joe found, but the original problem seems to be resolved.


kellyrob99:
+1 provided it works ;)

PR #1201, opened 2015-04-17
[NEXUS-8429] Rip out remaining user password "reset" bits from UI/backend

jdillon:
https://issues.sonatype.org/browse/NEXUS-8429

### DIFF
--- a/components/nexus-core/src/main/java/org/sonatype/nexus/internal/security/StaticSecurityConfigurationResourceImpl.groovy
+++ b/components/nexus-core/src/main/java/org/sonatype/nexus/internal/security/StaticSecurityConfigurationResourceImpl.groovy
@@ -13,13 +13,13 @@
 
 package org.sonatype.nexus.internal.security
 
+import javax.inject.Named
+import javax.inject.Singleton
+
 import org.sonatype.nexus.security.config.CPrivilege
 import org.sonatype.nexus.security.config.CRole
 import org.sonatype.nexus.security.config.MemorySecurityConfiguration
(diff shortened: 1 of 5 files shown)


kellyrob99:
+1

jdillon:
http://bamboo.s/browse/NX3-OSSF192-1 :white_check_mark:

components/nexus-core/src/main/java/org/sonatype/nexus/internal/security/StaticSecurityConfigurationResourceImpl.groovy
```diff
@@ -909,16 +909,19 @@ class StaticSecurityConfigurationResourceImpl
...
-                type: 'method',
-                name: 'User Reset Password - (delete,read)',
-                description: 'Give permission to reset any user\'s password.',
-                properties: [
-                    'method'    : 'delete,read',
-                    'permission': 'security:usersreset'
-                ]
-            ),
+
+            // NOTE: This privilege is no longer used by anything
```
adreghiciu: Why not just remove?

adreghiciu:
+1

components/nexus-core/src/main/java/org/sonatype/nexus/internal/security/StaticSecurityConfigurationResourceImpl.groovy
jdillon: I left here just to keep record of what 'id: 59' was.

I expect we'll need to overhaul the permissions at some time in the future to normalize namespace, etc and we can remove it then.

PR #1203, opened 2015-04-17
NEXUS-7898 ext model update

kellyrob99:
https://issues.sonatype.org/browse/NEXUS-7898

Notable here for conversation: @NotBlank versus @NotEmpty for Strings to prevent pure whitespace values from slipping in (http://stackoverflow.com/questions/17137307/in-hibernate-validator-4-1-what-is-the-difference-between-notnull-notempty)


kellyrob99:
:+1:  by CI http://bamboo.s/browse/NX3-OSSF191-1

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/store/AnalyticsEvent.js
```diff
@@ -37,6 +37,5 @@ Ext.define('NX.coreui.store.AnalyticsEvent', {
     }
   },
 
-  sortOnLoad: true,
```
kellyrob99: This property is already the default so I opted to remove it. Please speak up if you have a preference in this regard. :)

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/store/LdapServer.js
```diff
@@ -37,6 +37,5 @@ Ext.define('NX.coreui.store.LdapServer', {
     }
   },
 
-  sortOnLoad: true,
-  sorters: { property: 'name', direction: 'ASC' }
+  sorters: { property: 'order', direction: 'ASC' }
```
kellyrob99: opinionated changed based on the fact that the list has a User specified order

jdillon:
+1, though got a lot bigger than originally intended.  question about why all the sortOnLoad was removed.

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/store/AnalyticsEvent.js
jdillon: thx for clarification, no objection if this is the default

adreghiciu:
+1
Curious why did you add all of those constraints on XO's?

kellyrob99:
@adreghiciu 
``` Curious why did you add all of those constraints on XO's? ```
It seemed appropriate to add the annotations as I was identifying model properties where those constraints were already being applied elsewhere. It's still not 100% but IMO it's now a lot more representative of the domain.

PR #1204, opened 2015-04-18
[NEXUS-7981] Makes feed names consistent with their URLs

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-7981

@mosabua @jwayman input please.

@joedragons pointed out that feed names in NX3 are wildly inconsistent.

For example, there are two “New files” feeds. One has a URL of `…/recentlyCachedOrDeployedFiles` and the other is `…/recentlyChangedFiles`. Using “New” as a synonym for both “Cached or Deployed” and “Changed” seems confusing. We should use consistent terminology to describe things, and make sure feeds don’t have duplicate names.

This PR takes a stab at getting the names and URLs of feeds to use the same terminology, but more improvement could be made, and docs should probably lead that effort. I’m happy to facilitate in the implementation department. Let me know what you think.


jdillon:
+1

adreghiciu:
+1

cstamas:
+1

jwayman:
This change seems simple, straight-forward, and improved over what was generalized before. So, +1. I like it. 

As a side note, it would be interesting to see if users understand cached, deployed, and changed. Using my own experience, as a lay user/non-developer and novice Nexus user, I think have a definite understanding of "changed" (the thing it is now is not like it was before), I also think I understand "deployed" (this component has actually reached a state of use) - though I'm not sure to what extent - deployed in a build, a release, or something else. However, my inexperience with Nexus (perhaps development in general), makes cached the most ambiguous. I'm not sure what "cached" components are. I'd press on this a little bit as well to make sure that these concepts aren't unique to Maven (I don't think that's the case).

Finally, just to nag everyone about it... artifact vs. component should have a story(maybe it already does).

jdillon:
@jwayman we'll sort that out once we get equiv of this for CMA, ATM these are nearly all related to legacy repository framework events.  Lets not over polish this.

PR #1206, opened 2015-04-20
[NEXUS-8442] increase timeout for support-zip generation to 3 minutes

jdillon:
Also updates default ext.direct remoting provider to not retry and have 60 second timeout (instead of 30)

https://issues.sonatype.org/browse/NEXUS-8442

### DIFF
--- a/plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/support/SupportZip.js
+++ b/plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/support/SupportZip.js
@@ -35,6 +35,7 @@ Ext.define('NX.coreui.view.support.SupportZip', {
       xtype: 'nx-settingsform',
       settingsFormSubmitMessage: NX.I18n.get('ADMIN_SUPPORT_ZIP_CREATING'),
       settingsFormSuccessMessage: NX.I18n.get('ADMIN_SUPPORT_ZIP_CREATED'),
+      timeout: 60 * 3, // 3 minutes
       api: {
         submit: 'NX.direct.atlas_SupportZip.create'
       },
--- a/plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/app/Application.js
+++ b/plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/app/Application.js
@@ -229,9 +229,16 @@ Ext.define('NX.app.Application', {
    * Initialize Ex.Direct remote providers.
    */
   initDirect: function () {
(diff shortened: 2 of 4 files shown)


plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/support/SupportZip.js
```diff
@@ -35,6 +35,7 @@ Ext.define('NX.coreui.view.support.SupportZip', {
       xtype: 'nx-settingsform',
       settingsFormSubmitMessage: NX.I18n.get('ADMIN_SUPPORT_ZIP_CREATING'),
       settingsFormSuccessMessage: NX.I18n.get('ADMIN_SUPPORT_ZIP_CREATED'),
+      timeout: 60 * 3, // 3 minutes
```
jdillon: this is in seconds to be clear (timeout of provider is in millis) but form bits convert seconds to millis.

For ref:

http://docs-origin.sencha.com/extjs/4.2.3/#!/api/Ext.form.Panel-cfg-timeout
http://docs-origin.sencha.com/extjs/4.2.3/#!/api/Ext.form.Basic-cfg-timeout
http://docs-origin.sencha.com/extjs/4.2.3/#!/api/Ext.form.action.Action-cfg-timeout

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/app/Application.js
```diff
@@ -229,9 +229,16 @@ Ext.define('NX.app.Application', {
...
    */
   initDirect: function () {
-    var me = this;
+    var me = this,
+        remotingProvider;
+
+    remotingProvider = Ext.direct.Manager.addProvider(NX.direct.api.REMOTING_API);
+
+    // disable retry
+    remotingProvider.maxRetries = 0;
```
jdillon: Disabling retry, as otherwise in timeout, remoting provider will attempt the request again.  So in the case of support-zip (before this change) if it took longer than 30 seconds, then 2 support-zips will be created.

djsauble:
+1

kellyrob99:
+1, although we should probably keep an eye out in case the retry change impacts other areas of the UI

jdillon:
@kellyrob99 ftr I'm not sure what would use retry to start with, I was surprised to see it retry TBH.  I disabled this to avoid potential places where non-idempotent requests could cause more/strange problems.

jdillon:
http://bamboo.s/browse/NX3-OSSF198-1 :white_check_mark:

adreghiciu:
+1

PR #1211, opened 2015-04-21
NEXUS-8288 ui repo name validator

kellyrob99:
https://issues.sonatype.org/browse/NEXUS-8288


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/util/Validator.js
```diff
@@ -33,8 +34,42 @@ Ext.define('NX.util.Validator', {
   },
 
   /**
+   * @private
+   */
+  nxNameRegex : /^[^(_)#A-Z\\\/"'\*\?<>|\s,.][^#A-Z\\\/"'\*\?<>|\s,.]*$/,
```
kellyrob99: This should cover both valid characters for elastic search index, and URL. Please feel free to suggest better regex to accomplish the same task, and any other constraints we might like to enforce for names in our domain :)

kellyrob99:
@all If we agree on the pattern being applied here I will apply the same to the server java bean validation

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/util/Validator.js
mcculls: Note that this still accepts names containing reserved URL characters: http://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters, for example: "build&test". I assume these characters will be percent-encoded when we construct URLs based on such names?

kellyrob99: I'd rather just disallow them so we don't need to worry about it, but wanted to get some feedback before discounting that we might want to allow some and percent-encode others.
Anyone else feel strongly on this?

plugins/rapture/nexus-coreui-plugin/src/test/java/org/sonatype/nexus/coreui/RepositoryXOTest.groovy
```diff
@@ -0,0 +1,103 @@
...
+
+class RepositoryXOTest
+    extends Specification
+{
+
+  @Shared
+  Validator validator = Validation.buildDefaultValidatorFactory().validator
+
+  static List INVALID_CHARACTERS = ['_', '#', ',', 'A', '*', ' ', '#', '\'', '\\', '/', '?', '<', '>', '|', ' ', '\r',
+                                    '\n', '\t', ',', '.'].asImmutable()
```
jdillon: would it be simpler to have a set of valid chars instead of invalids?  /^[a-zA-z0-9._]+$/ ?

kellyrob99: If we expect anyone ever to use non-english characters, then no; although that presents its own problems. I opted to just negate characters that are 'known bad' either to elasticsearch for index naming, or for URLs(incomplete right now as @mcculls mentioned)

djsauble: I tend to agree with @jdillon. Whichever set is simplest to represent in an error message is probably the way to go. Long error messages are bad.

As far as I know, we have no plan to support non-english languages, but we probably should.

jdillon: I'm leaning towards making it simple and strict, since these are used as identifiers.  Don't care about non-english use of repository names.  We'll add a description field at some point in the future that folks can put whatever they want in, in any language, but the name should be rigid.

plugins/rapture/nexus-coreui-plugin/pom.xml
```diff
@@ -67,6 +67,12 @@
       <artifactId>goodies-testsupport</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.spockframework</groupId>
+      <artifactId>spock-core</artifactId>
+      <version>1.0-groovy-2.3</version>
+      <scope>test</scope>
+    </dependency>
```
jdillon: define deps in buildsupport please.

plugins/rapture/nexus-coreui-plugin/src/test/java/org/sonatype/nexus/coreui/RepositoryXOTest.groovy
```diff
@@ -0,0 +1,105 @@
...
+import javax.validation.Validation
+import javax.validation.Validator
+
+import org.sonatype.nexus.validation.group.Create
+
+import spock.lang.Shared
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class RepositoryXOTest
```
jdillon: javadocs always please

adreghiciu:
+1

kellyrob99:
:+1: by CI http://bamboo.s/browse/NX3-OSSF203-3

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/search/SearchServiceImpl.java
```diff
@@ -315,4 +317,13 @@ private String componentMetadata(final Component component, final Iterable<Asset
     return producer.getMetadata(component, assets);
   }
 
+
+  /**
+   * Sanitize repository name in a consistent fashion to ensure that the name used for an index is safe.
+   */
+  @NotNull
+  private String safeIndexName(final Repository repository) {
```
adreghiciu: I would just name it indexName()

adreghiciu:
+1

plugins/rapture/nexus-coreui-plugin/pom.xml
```diff
@@ -67,6 +67,24 @@
...
+
+    <dependency>
+      <groupId>org.spockframework</groupId>
+      <artifactId>spock-guice</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.sonatype.nexus</groupId>
+      <artifactId>nexus-extender</artifactId>
```
kellyrob99: @jdillon I threw this in strictly to keep the testing here, but wanted your feedback as to whether the test should move, or some alternative...

jdillon: @kellyrob99 I've no objection to spock, just never used it and I'm not yet sure how to read/write these things

kellyrob99: Docs have gotten much better with the 1.0 release: http://spockframework.github.io/spock/docs/1.0/
Couple of concrete notes:
- the when/then and other blocks are all Java labels, allowing the test runner to pre-process them before run; this can be a bit confusing because the code is not necessarily executed top-to-bottom
- where blocks essentially turn the method into the equivalent of a JUnit Parameterized test

kellyrob99: @jdillon my comment was actually more about nexus-extender here, which i dragged in to configure the Validator with guice

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/config/UniqueRepositoryNameValidatorTest.groovy
```diff
@@ -0,0 +1,47 @@
...
+package org.sonatype.nexus.repository.config
+
+import org.sonatype.nexus.repository.Repository
+import org.sonatype.nexus.repository.manager.RepositoryManager
+
+import spock.lang.Specification
+
+/**
+ * Tests validity of Repository names validated by {@link UniqueRepositoryNameValidator}
+ * @since 3.0
```
jdillon: not much point in since on tests or test classes, i tend to remove these anytime I see them.

plugins/rapture/nexus-coreui-plugin/src/test/java/org/sonatype/nexus/coreui/TestRepositoryManagerModule.groovy
```diff
@@ -0,0 +1,34 @@
...
+ */
+class TestRepositoryManagerModule
+    extends AbstractModule
+{
+  public static final NAMES = ['foo', 'bar', 'baz'].asImmutable()
+  
+  @Override
+  protected void configure() {
+    bind(RepositoryManager).toInstance({ -> NAMES.collect { { -> it } as Repository } } as RepositoryManager);
+  }
```
jdillon: ugh, so injected validator means we have to use injection for UT... thats not super desirable given we have spent a lot of time trying to remove use of injection from UT

kellyrob99: Technically we could do an end run around Hibernate here and provide a no-op impl of the custom validator through configuration of the Hibernate bits, but that would also preclude actually testing that the validation was applied...

plugins/rapture/nexus-coreui-plugin/pom.xml
jdillon: fair enough I didn't even notice that, certainly shouldn't be depending on extender here

mcculls: vote for https://github.com/sonatype/nexus-oss/pull/1222 - then you can depend on "nexus-validation" instead :)

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/config/UniqueRepositoryName.java
```diff
@@ -0,0 +1,42 @@
...
+ * Validate a name as unique across all Repositories.
+ *
+ * @since 3.0
+ */
+@Target({FIELD, METHOD, PARAMETER, ANNOTATION_TYPE})
+@Retention(RUNTIME)
+@Constraint(validatedBy = UniqueRepositoryNameValidator.class)
+public @interface UniqueRepositoryName
+{
+  String message() default "Name must be case-insensitively unique";
```
mcculls: sounds clunky, perhaps "Name must be unique (ignoring case)" ?

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/config/UniqueRepositoryNameValidator.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  public void initialize(final UniqueRepositoryName constraintAnnotation) {
+    // do nothing
+  }
+
+  @Override
+  public boolean isValid(final String value, final ConstraintValidatorContext context) {
+    for (Repository repository : repositoryManager.browse()) {
+      if (repository.getName().toLowerCase(Locale.ENGLISH).equals(value.toLowerCase(Locale.ENGLISH))) {
```
mcculls: FYI, Jason added `Strings2.lower(text)` as a shorthand in nexus-common

jdillon: After I saw this I was considering adding Strings2.equalsIgnoreCase(String,String) as well, since Strings.equalsIgnoreCase() doesn't have a locale option... fucking locale

components/nexus-validation/src/main/java/org/sonatype/nexus/validation/ValidationModule.java
```diff
@@ -45,16 +47,18 @@ protected void configure() {
...
 
   @Provides
   @Singleton
-  ValidatorFactory validatorFactory() {
+  ValidatorFactory validatorFactory(ConstraintValidatorFactory constraintValidatorFactory) {
     ClassLoader tccl = Thread.currentThread().getContextClassLoader();
     try {
       Thread.currentThread().setContextClassLoader(HibernateValidator.class.getClassLoader());
-
+
```
mcculls: minor: dangling whitespace

mcculls:
+1  🖖

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/config/UniqueRepositoryNameValidator.java
jdillon: Actually after talking to @mcculls this could/should just use Strings.equalsIgnoreCase() ?

kellyrob99: d3c40e2f4440fabc989aaa038e90f622b0047f08

PR #1214, opened 2015-04-21
[NEXUS-7719] Converts PEM and Certificate fields to monospace font

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-7719

Related: https://github.com/sonatype/nexus-pro/pull/525

This is a trivial change. Are there any other places in the UI which should be styled with monospace font?


adreghiciu:
+1

jdillon:
+1

@djsauble unsure if there are others, but we'll fix them as we find 'em I guess ;-)

PR #1228, opened 2015-04-23
StorageTx Fixes

mrprescott:
https://issues.sonatype.org/browse/NEXUS-8509

### DIFF
--- a/components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/Entity.java
+++ b/components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/Entity.java
@@ -23,6 +23,15 @@ public abstract class Entity
 {
   private EntityMetadata metadata;
 
+  /**
+   * Does the Entity correspond to a database record?
+   *
+   * This is not a dirtiness check.
+   */
+  public boolean isPersisted() {
+    return metadata != null;
+  }
+
   @Nullable
   public EntityMetadata getEntityMetadata() {
     return metadata;
--- a/components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/EntityId.java
+++ b/components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/EntityId.java
@@ -45,11 +45,11 @@ public class EntityId
       return true;
     }
 
-    EntityId entityId = (EntityId) o;
-
     if (!(o instanceof EntityId)) {
       return false;
     }
+    
+    EntityId entityId = (EntityId) o;
 
     if (!getValue().equals(entityId.getValue())) {
       return false;
--- a/components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/EntityId.java
+++ b/components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/EntityId.java
@@ -24,11 +24,11 @@ import static com.google.common.base.Preconditions.checkNotNull;
  * @since 3.0
  */
 public class EntityId
-  implements Serializable
+    implements Serializable
 {
   private static final long serialVersionUID = 1L;
 
-  private final String value;
+  private String value;
 
   public EntityId(final String value) {
     this.value = checkNotNull(value);
@@ -44,13 +44,10 @@ public class EntityId
     if (this == o) {
       return true;
     }
-    if (o == null || getClass() != o.getClass()) {
-      return false;
-    }
 
     EntityId entityId = (EntityId) o;
 
-    if (!value.equals(entityId.value)) {
+    if (!getValue().equals(entityId.getValue())) {
       return false;
     }
 
@@ -59,11 +56,11 @@ public class EntityId
 
   @Override
   public int hashCode() {
-    return value.hashCode();
+    return getValue().hashCode();
   }
 
   @Override
   public String toString() {
-    return value;
+    return getValue();
   }
 }
--- a/components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/EntityId.java
+++ b/components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/EntityId.java
@@ -28,7 +28,7 @@ public class EntityId
 {
   private static final long serialVersionUID = 1L;
 
-  private String value;
+  private final String value;
 
   public EntityId(final String value) {
     this.value = checkNotNull(value);
@@ -46,6 +46,10 @@ public class EntityId
     }
 
     EntityId entityId = (EntityId) o;
+
+    if (!(o instanceof EntityId)) {
+      return false;
+    }
 
     if (!getValue().equals(entityId.getValue())) {
       return false;
--- a/components/nexus-orient/src/main/java/org/sonatype/nexus/orient/entity/EntityAdapter.java
+++ b/components/nexus-orient/src/main/java/org/sonatype/nexus/orient/entity/EntityAdapter.java
@@ -174,7 +174,7 @@ public abstract class EntityAdapter<T extends Entity>
     checkNotNull(entity);
 
     // new entity must not already have metadata
-    checkState(entity.getEntityMetadata() == null);
+    checkState(!entity.isPersisted());
 
     ODocument doc = db.newInstance(typeName);
     return writeEntity(doc, entity);
@@ -309,7 +309,7 @@ public abstract class EntityAdapter<T extends Entity>
   /**
    * Return record identity of entity.
    */
-  protected ORID recordIdentity(final T entity) {
+  public ORID recordIdentity(final T entity) {
     return recordIdentity(id(entity));
   }
 
--- a/components/nexus-orient/src/main/java/org/sonatype/nexus/orient/entity/EntityAdapter.java
+++ b/components/nexus-orient/src/main/java/org/sonatype/nexus/orient/entity/EntityAdapter.java
@@ -205,6 +205,8 @@ public abstract class EntityAdapter<T extends Entity>
   {
     private final EntityAdapter owner;
 
+    private ODocument document;
+
     private final ORID rid;
 
     private final ORecordVersion rversion;
@@ -215,6 +217,7 @@ public abstract class EntityAdapter<T extends Entity>
 
     public AttachedEntityMetadata(final EntityAdapter owner, final ODocument document) {
       this.owner = checkNotNull(owner);
+      this.document = checkNotNull(document);
       checkNotNull(document);
       this.rid = document.getIdentity();
       this.rversion = document.getRecordVersion();
--- a/components/nexus-orient/src/main/java/org/sonatype/nexus/orient/entity/EntityAdapter.java
+++ b/components/nexus-orient/src/main/java/org/sonatype/nexus/orient/entity/EntityAdapter.java
@@ -12,6 +12,7 @@
  */
 package org.sonatype.nexus.orient.entity;
 
+import javax.annotation.Nonnull;
 import javax.inject.Inject;
 
 import org.sonatype.nexus.common.entity.Entity;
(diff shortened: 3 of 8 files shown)


components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/Entity.java
```diff
@@ -23,6 +23,15 @@
...
   private EntityMetadata metadata;
 
+  /**
+   * Does the Entity correspond to a database record?
+   *
+   * This is not a dirtiness check.
+   */
+  public boolean isPersisted() {
+    return metadata != null;
+  }
```
jdillon: Any reason this is not on EntityMetadata?

components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/EntityId.java
```diff
@@ -44,13 +44,10 @@ public boolean equals(final Object o) {
     if (this == o) {
       return true;
     }
-    if (o == null || getClass() != o.getClass()) {
-      return false;
-    }
```
jdillon: if you remove this, you must add an instanceof check instead to avoid equals() CCE

components/nexus-common/src/main/java/org/sonatype/nexus/common/entity/EntityId.java
```diff
@@ -24,11 +24,11 @@
  * @since 3.0
  */
 public class EntityId
-  implements Serializable
+    implements Serializable
 {
   private static final long serialVersionUID = 1L;
 
-  private final String value;
+  private String value;
```
jdillon: technically this can still be final

mrprescott:
http://bamboo.s/browse/NX3-OSSF218-2 :whitecheck:

components/nexus-orient/src/main/java/org/sonatype/nexus/orient/entity/EntityAdapter.java
```diff
@@ -259,33 +295,28 @@ protected void setMetadata(final T entity, final ODocument doc) {
...
   /**
    * Return record identity of entity.
    */
-  protected ORID recordIdentity(final T entity) {
-    EntityMetadata metadata = getMetadata(entity);
-    if (metadata instanceof AttachedEntityMetadata) {
-      return ((AttachedEntityMetadata) metadata).rid;
-    }
-    return getRecordIdObfuscator().decode(getType(), metadata.getId().getValue());
+  public ORID recordIdentity(final T entity) {
```
kellyrob99: Not directly related to the PR since this name was pre-existing, but anyone else think this should be prefixed with 'get' so it's clear that it's not an action? When I first saw the calls it was not clear this was a getter and not performing an action

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
```diff
@@ -347,18 +354,34 @@ public void roundTripTest() {
...
       asset2.name("asset2");
       tx.saveAsset(asset2);
+
       tx.commit();
 
+      // In transaction mode, ORIDs are placeholders until commit, so IDs should be collected after commit
+      asset1Id = id(asset1);
+      asset2Id = id(asset2);
+      componentId = id(component);
+      System.err.println(componentId);
```
kellyrob99: cruft

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
```diff
@@ -347,18 +354,34 @@ public void roundTripTest() {
...
+
+    try (StorageTx tx = underTest.openTx()) {
+      Bucket bucket = tx.getBucket();
+
       checkSize(tx.browseAssets(bucket), 2);
       checkSize(tx.browseComponents(bucket), 1);
 
-      assertNotNull(tx.findAsset(asset1.getEntityMetadata().getId(), bucket));
-      assertNotNull(tx.findComponent(component.getEntityMetadata().getId(), bucket));
+      // TODO: Remove test code
```
kellyrob99: cruft

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
kellyrob99: cruft

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
```diff
@@ -561,4 +604,160 @@ public void duplicateAssetName() throws Exception {
...
+      tx.commit();
+    }
+
+    try (StorageTx tx = underTest.openTx()) {
+      final Component component = tx.findComponentWithProperty("version", "testVersion", tx.getBucket());
+      final Iterable<Asset> assets = tx.browseAssets(component);
+      final Asset asset = assets.iterator().next();
+
+      assertThat(asset.formatAttributes().get("attribute1", String.class), equalTo("original"));
+      assertThat(asset.formatAttributes().get("attribute1", String.class), equalTo("original"));
```
kellyrob99: duplicate verification here, assuming you wanted to check attribute2

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
```diff
@@ -561,4 +604,160 @@ public void duplicateAssetName() throws Exception {
...
+  @Test
+  public void transactionsRollBackOnException() throws Exception {
+    try {
+      try (StorageTx tx = underTest.openTx()) {
+        final Component component = tx.createComponent(tx.getBucket(), testFormat)
+            .group("myGroup")
+            .version("0.9")
+            .name("myComponent");
+        tx.saveComponent(component);
+        if (true) {
```
kellyrob99: why the if check?

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
mrprescott: ack

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
mrprescott: Muscle memory to avoid an unreachable code exception, but it's the last line in the block. Removed.

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
```diff
@@ -561,4 +604,160 @@ public void duplicateAssetName() throws Exception {
...
+          throw new MidTransactionException();
+        }
+      }
+    }
+    catch (MidTransactionException ignored) {
+    }
+
+    try (StorageTx tx = underTest.openTx()) {
+      final Component component = tx.findComponentWithProperty("group", "myGroup", tx.getBucket());
```
kellyrob99: I think this test is a bit bogus, we get the same result regardless of the thrown exception...

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
```diff
@@ -561,4 +604,160 @@ public void duplicateAssetName() throws Exception {
...
+          .version("0.9")
+          .name("myComponent");
+      tx.saveComponent(component);
+      tx.commit();
+    }
+
+    try (StorageTx tx = underTest.openTx()) {
+      final Iterable<Component> components = tx.browseComponents(tx.getBucket());
+      for (Component c : components) {
+        System.err.println(c);
```
kellyrob99: cruft

mrprescott: already gone

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
```diff
@@ -561,4 +604,160 @@ public void duplicateAssetName() throws Exception {
...
+
+      final Component component = tx.findComponentWithProperty("group", "myGroup", tx.getBucket());
+      assertThat(component, is(notNullValue()));
+      assertThat(component.group(), is("myGroup"));
+    }
+  }
+
+  @Test
+  public void entityIdCanBeUsedInLaterTransactions() throws Exception {
+    EntityId componentId = null;
```
kellyrob99: no need for explicitly setting null, and then IDEA will stop flagging them too :)

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
```diff
@@ -561,4 +604,160 @@ public void duplicateAssetName() throws Exception {
...
+
+    try (StorageTx tx = underTest.openTx()) {
+      final Component component = tx.findComponent(componentId, tx.getBucket());
+      assertThat("component", component, is(notNullValue()));
+      assertThat(component.name(), is("component"));
+    }
+  }
+
+  @Test
+  public void multipleEntityIdsCanBeUsedInLaterTransactions() throws Exception {
```
kellyrob99: this covers the previous test conditions completely afaict, might as well just rename to 'entityIdCanBeUsedInLaterTransactions' and delete that present implementation

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
mrprescott: I've got mixed feelings about that. I agree, if that exception wasn't thrown (and in the previous test, if rollback wasn't called), the effect would be the same. However, in a sense the test is there to demonstrate a capability (one that must be considered if, say, we decided to go with automatic committing at some point).

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
```diff
@@ -561,4 +604,160 @@ public void duplicateAssetName() throws Exception {
...
+      assertThat("component", component, is(notNullValue()));
+      assertThat(component.name(), is("component"));
+
+      final Asset asset = tx.findAsset(assetId, tx.getBucket());
+      assertThat("asset", asset, is(notNullValue()));
+      assertThat(asset.name(), is("hello"));
+    }
+  }
+
+  private static class MidTransactionException
```
kellyrob99: really no need for custom exception here, is there? seems like overkill, even if it's just a couple of lines

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
```diff
@@ -561,4 +604,160 @@ public void duplicateAssetName() throws Exception {
...
+      extends Exception
+  {}
+
+  @Test
+  public void dependentQueryFromNewComponent() throws Exception{
```
kellyrob99: incomplete?

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
mrprescott: Yes, possibly. There was a point in my restoration of transactions and id nonsense where the first test passed but the second didn't. I suppose a better name would be 'entityIdsCanBeReferencedBeforeCommit'.

components/nexus-repository/src/test/java/org/sonatype/nexus/repository/storage/StorageFacetImplIT.java
mrprescott: No, but it needs an explanatory comment.

PR #1232, opened 2015-04-23
NEXUS-8128 - Persist results columns on further actions

kellyrob99:
~~- initial prototype for universally setting stateful on Master views~~
After discussion with @jdillon , just configuring each view on its own.

https://issues.sonatype.org/browse/NEXUS-8128


kellyrob99:
This would set all of our Master views to store state by default, and would require any child impl's that didn't want this behaviour to override settings(most likely in initComponent() override?).
In order to get more than simple sort ordering stored, we need an explicit id on all columns - including those not allowed for configuration like the 'drilldown affordance' column - otherwise the correct column order can't be re-established.

Alternatively, we could configure statefulness only on the grids we want to support. The downside of that is a lack of consistency.
If one of our more experienced ExtsJS devs has a better way to apply a config option like this without explicitly configuring each view, please share :)  Keep in mind that we need a recognizable unique ID per component type, which maps quite nicely in this case to the alias of each Master child class.

If anyone wants to play with this, you can see what gets stored with an invocation in the console like:
```
Ext.state.Manager.get('nx-coreui-repository-list')
```
... with the appropriate widget alias of course.

djsauble:
I can’t imagine a scenario where we would **not** want grid lists to have state awareness, so +1 on the blanket approach.

As far as a better way to set `id`, perhaps the following code in the Master.initComponent function?

    Ext.each(me.columns, function(col) {
        if (col.dataIndex) {
            col.id = col.dataIndex;
        }
    });

That should make it dynamic. I haven’t actually checked whether all master views use dataIndex though.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/drilldown/Master.js
```diff
@@ -29,6 +33,15 @@ Ext.define('NX.view.drilldown.Master', {
   initComponent: function() {
     var me = this;
 
+    if(!Ext.isEmpty(me.alias) && !Ext.Array.contains(me.alias, 'widget.nx-drilldown-master')) {
+      var widgetId = me.alias[0].replace(/^(widget\.)/, ""); // given possibly multiple aliases, just pick the first
+      me.logDebug('Setting stateful under stateId: ' + widgetId);
+      Ext.apply(me, {
+        stateful: true,
+        stateId: widgetId
+      });
```
jdillon: seems kinda useless waste of call to Ext.apply

jdillon:
Hrm... why not just set the stateful/stateId in the classes like is done in the other places we do this?

The system already has enough magic/dynamic stuff... why do we need more for something that is otherwise pretty simple and straight-forward?

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/repository/RepositoryList.js
```diff
@@ -32,12 +32,13 @@ Ext.define('NX.coreui.view.repository.RepositoryList', {
...
+      id: 'icon'
     },
-    { header: NX.I18n.get('ADMIN_REPOSITORIES_LIST_NAME_COLUMN'), dataIndex: 'name', flex: 1 },
-    { header: NX.I18n.get('ADMIN_REPOSITORIES_LIST_TYPE_COLUMN'), dataIndex: 'type' },
-    { header: NX.I18n.get('ADMIN_REPOSITORIES_LIST_FORMAT_COLUMN'), dataIndex: 'format' },
-    { header: NX.I18n.get('ADMIN_REPOSITORIES_LIST_STATUS_COLUMN'), dataIndex: 'status', flex: 1, 
+    { header: NX.I18n.get('ADMIN_REPOSITORIES_LIST_NAME_COLUMN'), dataIndex: 'name', id: 'name', flex: 1 },
+    { header: NX.I18n.get('ADMIN_REPOSITORIES_LIST_TYPE_COLUMN'), dataIndex: 'type' , id: 'type' },
+    { header: NX.I18n.get('ADMIN_REPOSITORIES_LIST_FORMAT_COLUMN'), dataIndex: 'format', id: 'format' },
+    { header: NX.I18n.get('ADMIN_REPOSITORIES_LIST_STATUS_COLUMN'), dataIndex: 'status', id: 'status', flex: 1,
```
jdillon: why do we need 'id'?  isnt' that the globally unique component id?

http://docs-origin.sencha.com/extjs/4.2.3/#!/api/Ext.AbstractComponent-cfg-id

unrelated, looks like 'header' is deprecated and should be 'text' instead:

http://docs-origin.sencha.com/extjs/4.2.3/#!/api/Ext.grid.column.Column-cfg-header

kellyrob99: id is used as the column key when storing, without it you get synthetic 'header-1234' auto-generated ids written to local storage, which is of no use whatsoever

kellyrob99:
@jdillon We have a couple dozen grids like this so I was aiming for something more consistent, and so we don't end up with C&P or having to remember to set this for future impl's. I'm thinking now that the best answer might be a grid mixin that also exposes a method to clear the state easily. Whether that gets applied globally to all Master views or individually, it should at least encapsulate any customizations we need to do.

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/repository/RepositoryList.js
jdillon: we may have to look at this closer then, as id is globally unique to the entire app.  So you are saying the id='name' component for the entire app is a column.

Looks like columns have a stateId that should be used for what you are attempting:

http://docs-origin.sencha.com/extjs/4.2.3/#!/api/Ext.grid.column.Column-cfg-stateId

jdillon:
@kellyrob99 my gut says the UI code has already started to grow to the point of complexity that its hard to comprehend, and away from standard use of extjs.  If we are talking about 2 extra fields in view impls to wire up the stateful configuration, that is IMO better than magic (under the covers how does it work who can remember) sort of change.  Lets not make this more complex than it needs...

jdillon:
@kellyrob99 btw originally I think of perhaps using a mixin to configure state, but that seemed more complex than just setting a boolean and a string.  Also wasn't clear how a mixin with other mixins behave since stateful functionality is also provided by a mixin: http://docs-origin.sencha.com/extjs/4.2.3/source/Stateful.html#Ext-state-Stateful

jdillon:
@kellyrob99 id/stateId for column that seems pretty odd that it can't be smarter about figuring out the right id to use :-(

default looks like its:

```
    getStateId: function () {
        return this.stateId || this.headerId;
    },
```

Anyways again I think this is cost of doing business and playing by the rules of standard extjs, so probably for now the simplest option is to define stateId on columns in grids that are stateful.

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/repository/RepositoryList.js
kellyrob99: ack, sorry I missed that

mrprescott:
+1

kellyrob99:
:+1: by CI http://bamboo.s/browse/NX3-OSSF222-1

kellyrob99:
Merging based on direct +1 on HC from @jdillon

PR #1238, opened 2015-04-24
Repository config forms: styling and content flow improvements

djsauble:
Before:
screen shot 2015-04-24 at 10 10 04 am

After:
screen shot 2015-04-24 at 10 15 12 am


kellyrob99:
+1 incrementally better, with the expectation that we'll likely continue to refine form layout going forward(here and globally)

PR #1241, opened 2015-04-24
[NEXUS-8206] Clears task type filter on navigate/destroy

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8206

The “Create task” button can become disabled even when there are task types available.

This happens because of the task type filter. When all entries are filtered out, Ext JS disables the create button, even if we navigate away or destroy the view.

This PR ensures we clear the filter in both of those cases.


adreghiciu:
-1
We should not clear all filters but only the filter related to the filtering box. So, I suggest that the filtering plugin will call into Filtering#clearFilter() on its destroy

kellyrob99:
I'm not sure that destroy will work here without additional changes, as it appears the filterbox is (at least presently) not being destroyed but just removed from the visible window

djsauble:
I moved the task-specific filtering logic to the destroy function of the filtering plugin, as @adreghiciu suggested.

However, I left the drilldown-specific filtering logic as is. If you use the breadcrumb to navigate backwards, we want to clear the filter for all grids forward in the hierarchy (since they are no longer linked in the breadcrumb). We’re already [resetting forms](https://github.com/sonatype/nexus-oss/pull/1241/files#diff-cf2dcfdfef6cc16afa11050295b674e2R269) in these panels, why not reset the grid filters too?

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/drilldown/Drilldown.js
```diff
@@ -273,6 +274,15 @@ Ext.define('NX.view.drilldown.Drilldown', {
           }
         });
       }
+
+      // Clear successive filters
+      if (i > index) {
```
kellyrob99: might as well put this in the same if block as above, make it clear that both are triggered by the same condition

kellyrob99:
+1 if it works, but we should keep an eye out for any interactions that don't comply with the scenario we're building here. It's possible that specific Drilldown interactions might not behave as expected when they're sharing singleton Stores, although I can't point at a specific example now.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/drilldown/Drilldown.js
djsauble: Done.

adreghiciu:
+1

djsauble:
Manual merge. https://github.com/sonatype/nexus-oss/commit/582f862c3f03b60625a8efa4e5f123783781f9ba

PR #1248, opened 2015-04-27
[NEXUS-8495] Make write policy mandatory for hosted repositories

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8495

The fix is to make write policy mandatory for hosted repos and set the value for default repositories


mrprescott:
+1

kellyrob99:
+1

PR #1250, opened 2015-04-27
[NEXUS-8493] Default timeout values for proxy repositories

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8493

The fix changes the config to use seconds instead of milliseconds, changes the defaults to 3 retries / 20 seconds timeout (this will be changed when we will take the defaults from global http config) and removes the default configuration for maven/nuget proxies.


cstamas:
+1 (btw I spotted this today, that maven2 timeout was too low, when started proxying RAO ;) !

kellyrob99:
+1

PR #1251, opened 2015-04-27
NEXUS-8502 password optional

kellyrob99:
https://issues.sonatype.org/browse/NEXUS-8502
http://bamboo.s/browse/NX3-OSSF241-1


jdillon:
-1 unless there is a valid reason why password should be optional, I think it should be required.

adreghiciu:
I looked into http client and password is not a required field (user name is)

kellyrob99:
Closing for now, will reopen with updated impl to ensure that passwords are required and serialized correctly as per https://issues.sonatype.org/browse/NEXUS-8511

djsauble:
I agree that not being required to set a password looks strange (even if legal).

We could make this behavior clearer by having a “do not set a password” checkbox. If checked, the password field is disabled. If not, the password field can be changed and submitted as normal.

Not for this ticket, obviously, but if there are any other instances in the UI where a password is optional, let’s consider this approach.

PR #1252, opened 2015-04-28
[NEXUS-8489] Bring back ssl certificate details window

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8489

The certificate window was transformed to a panel and used as part of drill down.


kellyrob99:
+1

djsauble:
+1

PR #1256, opened 2015-04-29
[NEXUS-8441] Mask entire screen when marking log

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8441

### DIFF
--- a/plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/logging/LogMark.js
+++ b/plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/logging/LogMark.js
@@ -35,9 +35,6 @@ Ext.define('NX.coreui.view.logging.LogMark', {
     defaults: {
       anchor: '100%'
     },
-    api: {
-      submit: 'NX.direct.logging_Log.mark'
-    },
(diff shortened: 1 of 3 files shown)


plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/logging/LogMark.js
```diff
@@ -46,7 +43,8 @@ Ext.define('NX.coreui.view.logging.LogMark', {
         xtype: 'textfield',
         name: 'message',
         itemId: 'message',
-        fieldLabel: NX.I18n.get('ADMIN_LOG_VIEWER_MARK_MESSAGE')
+        fieldLabel: NX.I18n.get('ADMIN_LOG_VIEWER_MARK_MESSAGE'),
+        allowBlank: false
```
jdillon: if you do not allow blank  you probably also need to blur on validation to avoid validation on cancel.

adreghiciu: It is not required because the validation messages is already shown, so pressing cancel will directly close the window.

jdillon:
+1

PR #1259, opened 2015-04-30
NEXUS-8157 timeout and disconnect

kellyrob99:
https://issues.sonatype.org/browse/NEXUS-8157
http://bamboo.s/browse/NX3-OSSF248-1

### DIFF
--- a/plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/UiSessionTimeout.js
+++ b/plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/UiSessionTimeout.js
@@ -24,7 +24,8 @@ Ext.define('NX.controller.UiSessionTimeout', {
     'NX.Messages',
     'NX.Security',
     'NX.State',
-    'NX.I18n'
+    'NX.I18n',
+    'NX.State'
   ],
   mixins: {
     logAware: 'NX.LogAware'
@@ -32,6 +33,13 @@ Ext.define('NX.controller.UiSessionTimeout', {
 
   views: [
     'ExpireSession'
+  ],
+  
+  refs: [
+    {
+      ref: 'expireSessionWindow',
+      selector: 'nx-expire-session'
+    }  
   ],
 
   SECONDS_TO_EXPIRE: 30,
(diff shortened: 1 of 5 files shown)


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/User.js
```diff
@@ -248,10 +248,14 @@ Ext.define('NX.controller.User', {
           win.options.success.call(win.options.scope, win.options);
         }
       },
-      failure: function() {
+      failure: function(response) {
+        var message = NX.I18n.get('GLOBAL_SERVER_INCORRECT_CREDENTIALS_WARNING');
```
kellyrob99: opinionated change here as disconnect would still allow signin attempts, which was hard-wired to tell you that you had wrong user/pass

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/UiSessionTimeout.js
```diff
@@ -55,30 +64,16 @@ Ext.define('NX.controller.UiSessionTimeout', {
...
-
-  /**
-   * @private
-   */
-  onUserChanged: function (user) {
-    var me = this,
-        uiSettings = NX.State.getValue('uiSettings') || {};
-
-    me.setupTimeout(user ? uiSettings['sessionTimeout'] : undefined);
+    this.setupTimeout();
```
djsauble: Minor, but convention is to use `me` instead of `this` directly.

kellyrob99: I don't see the point of introducing the temp variable me for the single call required, but totally agree that it makes js scoping+readability better when we need to access it more than once or in nested code blocks.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/User.js
djsauble: Nice find. I’ve been bothered by that behavior before.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/UiSessionTimeout.js
djsauble: It’s in the same spirit as always adding curly brackets to code, even when not strictly necessary, to lay the groundwork for future complexity. But I could go either way.

djsauble:
+1

adreghiciu:
+1

PR #1262, opened 2015-05-04
slight optimization by avoiding duplicate File.lastModified()

peterlynch:
slight optimization by avoiding duplicate File.lastModified()


cstamas:
+1

mrprescott:
+1

kellyrob99:
+1

peterlynch:
I did not measure what gain this has, but seemingly we are going to native code 2x less frequently than before.

Noticed in a customer thread dump for a heavily loaded Nexus a thread that was checking last modified in this area.

PR #1267, opened 2015-05-05
NEXUS-8552 - Enable ExtJS Diagnostics for 4.x layout issues

kellyrob99:
- hide the storage fieldset to prevent layout failure
- enable diagnostics which were useful to find and fix this otherwise obtuse problem

https://issues.sonatype.org/browse/NEXUS-8552

### DIFF
--- a/plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/repository/facet/StorageFacet.js
+++ b/plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/repository/facet/StorageFacet.js
@@ -40,6 +40,7 @@ Ext.define('NX.coreui.view.repository.facet.StorageFacet', {
         xtype: 'fieldset',
         cls: 'nx-form-section',
         title: NX.I18n.get('ADMIN_REPOSITORIES_DETAILS_SETTINGS_STORAGE_FACET'),
+        hidden: true,
 
         items: [
           // TODO add blob store selection
--- a/plugins/rapture/nexus-rapture-baseapp/src/main/resources/static/rapture/ext/src/diag/layout/Context.js
+++ b/plugins/rapture/nexus-rapture-baseapp/src/main/resources/static/rapture/ext/src/diag/layout/Context.js
@@ -1,0 +1,540 @@
+/*
+This file is part of Ext JS 4.2
+
+Copyright (c) 2011-2014 Sencha Inc
+
+Contact:  http://www.sencha.com/contact
+
+Commercial Usage
+Licensees holding valid commercial licenses may use this file in accordance with the Commercial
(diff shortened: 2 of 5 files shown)


plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/repository/facet/StorageFacet.js
```diff
@@ -40,6 +40,7 @@ Ext.define('NX.coreui.view.repository.facet.StorageFacet', {
         xtype: 'fieldset',
         cls: 'nx-form-section',
         title: NX.I18n.get('ADMIN_REPOSITORIES_DETAILS_SETTINGS_STORAGE_FACET'),
+        hidden: true,
```
kellyrob99: here is the 'fix' without which the layout blew up with the empty fieldset, an obscure error related to 'shrinkwrap'

djsauble:
+1

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/repository/facet/StorageFacet.js
djsauble: Good find. Is the lack of fields in this facet temporary?

jdillon:
what are these new js classes and where did they come from?

kellyrob99:
@jdillon these are extjs classes which add detailed logging for UI layout, as explained in the link above ^^^

jdillon:
can we fix the bug and hten leave the addition of diag stuff for consideration later?  i'm not sure we want to get into the business of checking in specific stuff like this w/o a bit more consideration for how we'll manage it.  things like this likely make moving to extjs5 even harder, so we want to be very careful.

kellyrob99:
@jdillon I am fine with that, will leave this PR as POC in case this comes up, as without this in place the alternative to chase down issues like this appears to be commenting out components one at a time until you find the one with the problem.

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/repository/facet/StorageFacet.js
kellyrob99: Yes, at some point in the future we will want to be able to configure which backing BlobStore is used by a repository, but probably not for awhile.

kellyrob99:
Merged the single-line fix in 604a1e7c832d1e97c8cb1f0fc3141b12be33f32d

kellyrob99:
For posterity's sake, I checked and these files are not present in 5.x source code anyhow so @jdillon 's point is... right on point

PR #1269, opened 2015-05-07
[NEXUS-7711] Make the index/describe pages consistent with the rest of the UI

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-7711

#### Before

screen shot 2015-05-07 at 9 08 26 am

#### After

screen shot 2015-05-07 at 10 27 19 am

#### Before

screen shot 2015-05-07 at 10 56 20 am

#### After

screen shot 2015-05-07 at 10 50 25 am


kellyrob99:
Processing time shown in the 'after' version of Describe contains a question mark, not sure if that's a problem with code changed here or elsewhere.

kellyrob99:
+1 on the UI L&F changes, but should target CMA repositories instead of the legacy endpoints which are either already gone or soon to be.

djsauble:
'?' behavior was in the old code as well. First load is fine (shows `ms`), but subsequent refreshes have the question mark (shows `?s`). Not sure why. Perhaps @mrprescott knows why?

djsauble:
@kellyrob99 Won’t the CMA endpoints point to the same templates? Seems like it should Just Work™.

kellyrob99:
'?' is a rendering issue with the underlying stopwatch impl https://github.com/google/guava/issues/844 
I will raise a separate issue to look at that

adreghiciu:
+1

jdillon:
@kellyrob99 @djsauble the '?' problem is indeed due to guava using unicode chars for micro symbol and the string encoding for html is translating this to an undisplayable char.  I've addressed this in: https://github.com/sonatype/nexus-oss/pull/1273, but we may want to eventually consider something smarter overall (perhaps not using velocity which has many warts and clungy syntax)

jdillon:
should probably figure out where this will live, before merging so we don't loose this

djsauble:
@jdillon I noticed that `nexus-content-plugin` has been deprecated. The describe template was migrated to `nexus-repository-httpbridge`, so I ported my changes over.

However, the content template is nowhere to be found. Has it been rewritten yet?

djsauble:
@mcculls confirmed that `repositoryContentHtml.vm` does not have a replacement yet. I’ve moved the old template into the `nexus-repository-httpbridge` plugin for now.

When a replacement is written, I’ll port the old styles and delete it. Seem reasonable?

jdillon:
+1 is fine, will sort out index.html template when/if we need it, for now its fine to leave it where it is, though unused.

PR #1270, opened 2015-05-07
NEXUS-8503 - Maven "Version policy" uneditable but shows as normal text field would

kellyrob99:
- Styling rules for disabled combobox, consistent with other readonly fields
- Reverts “nx-no-border” style for simplicity. We should have only one style for readonly fields.

Issue: https://issues.sonatype.org/browse/NEXUS-8503

screen shot 2015-05-11 at 10 43 43 am


plugins/rapture/nexus-rapture-baseapp/src/main/baseapp/packages/rapture-theme/sass/src/Ext/form/field/Base.scss
```diff
@@ -8,10 +8,25 @@
  * cls: 'x-form-readonly'
  */
 
-.x-form-field[readonly].x-form-field:not([role="combobox"]) {
+@mixin nx-form-field-readonly {
   color: $color-dark-gray;
-  background-color: $color-smoke;
```
djsauble: Why did you remove the background color from read-only fields? Seems like a regression.

kellyrob99: Depends on what we want these fields to look like. The only readonly fields I am aware of are on this page, and they all also apply nx-no-border which immediately overwrites this background. If we want to be consistent with readonly fields on other forms, they should also use the nx-no-border styling, so we might want to just merge the two rules here instead.

djsauble: Most of the predefined roles have read-only fields.

I think there’s a difference between an immutable field and a field which you don’t have permissions to edit (or can’t due to some other condition). I’m okay with immutable fields getting the nx-no-border treatment (since they are, essentially, labels), but I think all other readonly fields need to look like input boxes.

If a readonly input field doesn’t have a grey background, it looks the same as an empty field with placeholder text, which is the reason I added a smoke background.

roles - sonatype nexus 2015-05-08 17-11-37

kellyrob99: Makes sense, thanks for the explanation, will put back this styling rule.

kellyrob99:
After discussion with @djsauble we're going to go another direction for the sake of consistency

adreghiciu:
+1

PR #1275, opened 2015-05-08
[NEXUS-8542] Remove create panels when sliding panels

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8542

The problem is that editable conditions are based on the fact that there is only one form of a certain type active when the edit panel is shown. yet, the add panels are still there but hidden, so the fix will remove the ones after the current shown one.


mrprescott:
+1

djsauble:
+1

PR #1280, opened 2015-05-08
[NEXUS-8525] Show/Hide NTLM auth fields on change of combo

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8525


kellyrob99:
Obviously still some extjs subtleties I need to wrap my head around :)
+1

PR #1283, opened 2015-05-08
NEXUS-8576 - Cannot submit tasks with advanced (cron) filled in

kellyrob99:
- include cronExpression in form submit
- ensure that rejected cronExpression is matched to the right field, otherwise it shows as a transient popup that is easy to miss

https://issues.sonatype.org/browse/NEXUS-8576
http://bamboo.s/browse/NX3-OSSF271-1

Regex validated at submit:
screen shot 2015-05-08 at 4 52 23 pm

Quartz fails due to a logical constraint:
screen shot 2015-05-08 at 4 52 35 pm

### DIFF
--- a/plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/task/TaskSettingsForm.js
+++ b/plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/task/TaskSettingsForm.js
@@ -94,8 +94,8 @@ Ext.define('NX.coreui.view.task.TaskSettingsForm', {
     if (task.startDate) {
       task.startDate = task.startDate.toJSON();
     }
-    
-    if(task.schedule === 'advanced') {
+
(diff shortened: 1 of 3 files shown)


plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/task/TaskSettingsForm.js
```diff
@@ -94,6 +94,10 @@ Ext.define('NX.coreui.view.task.TaskSettingsForm', {
     if (task.startDate) {
       task.startDate = task.startDate.toJSON();
     }
+    
+    if(task.schedule === 'advanced') {
```
jdillon: please watch formatting

jdillon:
+1 if this fixes it, though would still like to know if we can use bean validation instead of validation-reposnse.

adreghiciu:
+1

PR #1290, opened 2015-05-11
Fix & clean SSL plugin

adreghiciu:
* merge rest resources used as services into ui components
* replace trust store keys with storing the "useTrustStore" in repository/smtp/ldap/clm/crowd configurations, such dropping a lot of code that was about storing this setting as a capability


adreghiciu:
Closing in favour of a different branch

PR #1301, opened 2015-05-13
[NEXUS-8538] Show non proxy hosts under http/https settings

adreghiciu:
… if http settings are enabled

https://issues.sonatype.org/browse/NEXUS-8538


djsauble:
+1

PR #1302, opened 2015-05-13
[NEXUS-8488] Recreate repository http client when global http config change

adreghiciu:
Somehow related https://issues.sonatype.org/browse/NEXUS-8488


cstamas:
+1

jdillon:
any implications of state corruption?

adreghiciu:
When facet is stopped (e.g. reconfiguration), the facet is unsubscribed so there will be no events received

jdillon:
+1

PR #1305, opened 2015-05-13
[NEXUS-8595] Clear filter on slide panel is redundant

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8595

During a drilldown transition, we delete any wizard panels which fall immediately after the current panel. This automatically cleans up the filter for the models backing those panels, so doing it manually here is redundant.


adreghiciu:
+1

djsauble:
Manual merge. https://github.com/sonatype/nexus-oss/commit/9d2280f3482403e94e334515fbb8c78cc1442122

PR #1308, opened 2015-05-14
Refactor search service

cstamas:
SearchService "silently" depended on fact that Repository on who's behalf indexes, does have StorageFacet attached. On the other hand, the SearchFacet was working/processing the StorageFacet related entities (Component). In one way, both of search related components were "lurking around" StorageFacet and related classes.

This refactor now makes SearchService completely independent of SearchFacet being attached to a Repository, as all the storage related work is now moved to a SearchFacet that IS now dependant on StorageFacet.

One question pending: use of EntityId in SearchFacet interface, as currently we have no other means to identify a Component entity.

CI
http://bamboo.s/browse/NX3-OSSF302


mrprescott:
+1 Don't hold up this PR because of the EntityID issue, this PR neither worsens it nor solves it. Let's tackle that separately.

PR #1310, opened 2015-05-14
[NEXUS-8620] Do not disable readonly combos in the repo config UI

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8620

Disabled combos are excluded from the values when a form is submitted. This doesn’t work for repo config, since the server assumes those values are present. This PR achieves the same readonly styling for combos without disabling them.


adreghiciu:
+1

PR #1312, opened 2015-05-14
[NEXUS-8417] Load the realm settings form after the realm type store has loaded

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8417


adreghiciu:
+1

PR #1257, opened 2015-04-29
NEXUS-8305: Rebuild Maven2 repository metadata

cstamas:
Rebuild maven metadata, replaces
https://github.com/sonatype/nexus-oss/pull/1237

Changes:
- rebuild Maven2 metadata related classes (3) + tests + IT
- API bits moved out of `internal` package
- introduced `MavenHostedFacet` (that exposes `rebuildMetadata()` single method for now)
- `MavenPath` improvements: for timestamped snapshot paths it extracts the timestamp and buildNumber too + locate method + bugfix
- `MavenFacet` stores baseVersion always along with other format specific attributes

**IMPORTANT**: SearchFacet calls "peppered" over MavenFacet (courtesy of Kelly) are removed for now, due to requirement to call them _after_ commit. This is temporary, PR for that pending.

Related:
- to return indexing https://github.com/sonatype/nexus-oss/pull/1289 

Issue
https://issues.sonatype.org/browse/NEXUS-8305

CI
http://bamboo.s/browse/NX3-OSSF246

PR #1319, opened 2015-05-15
NEXUS-8305: Maven2 metadata rebuild (Part3)

cstamas:
Splitting original PR #1257 into multiple smaller PR "steps" for easier review. This is step 3, with changes:
* introduced MavenHostedFacet (that exposes rebuildMetadata() single method for now)
* MavenFacet received "batch" methods (those explicitly using passed in TX, get/put/delete)
* MetadataRebuilder does the "component crawling"
* IT for rebuild

Related:
* to return indexing #1289 (already merged)

Classname and package cleanup pending, will be new PR (as Kelly pointed out, a lot of `/org/sonatype/maven2/Maven2MavenFooBar` classes overall in maven plugin)

Issue

CI
http://bamboo.s/browse/NX3-OSSF311


cstamas:
http://bamboo.s/browse/NX3-OSSF311-1 :+1:

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/MavenHostedFacet.java
```diff
@@ -0,0 +1,38 @@
...
+ * Maven hosted facet, present on all Maven hosted-type repositories.
+ *
+ * @since 3.0
+ */
+@Facet.Exposed
+public interface MavenHostedFacet
+    extends Facet
+{
+  /**
+   * Rebuilds/updates Maven metadata.
```
mrprescott: The intf could probably stand to be clearer about what exactly triggers replacing. Is 'update [not replace] everything' an impossible/undesirable operation?

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/metadata/MetadataRebuilder.java
```diff
@@ -0,0 +1,476 @@
...
+              }
+          )
+      ).execute(sqlParams);
+    }
+
+    /**
+     * Process exits from group level, executed in isolation.
+     */
+    private void rebuildMetadataExitGroup(final String currentGroupId) {
+      isolate(new Runnable()
```
mrprescott: So, something's not sitting right with the isolate method.  You're running a long, batch process that can't follow StorageTx's approach to transactions, but yet you need the helper methods build into StorageTx.  Essentially, StorageTx does two jobs - transaction control, and provides a suite of domain model manipulation tools (e.g. manipulating Components and Assets).

The isolate() method is necessary because of Orient DB's underlying addiction to thread-locals, and you're trying to use some of the functionality of StorageTx without being impacted by its assumption that it can control Orient.

Rather than pepper our code with this sort of suppression/hack stuff, I think we ought to tease apart the transaction control and domain-manipulation features of StorageTx - or perhaps, encapsulate the way that we're obtaining connections/transactions from Orient into a service that StorageTx also uses, to allow for this sort of isolation globally.

Anyways, let's not hold up this PR with this, but queue up another issue to deal with soon.

mrprescott:
+1

cstamas:
Merged as 56d55f4d33ac84eff1f3fbdf6384b7f2de3e2214

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/maven2/metadata/MetadataRebuilder.java
cstamas: Interesting experiment with working outcome: no need for multiple DBs, nested TXes, nor thread-local monkeying: https://github.com/sonatype/nexus-oss/pull/1327

PR #1320, opened 2015-05-15
NEXUS-8637: Handle illegal operation

cstamas:
and make a sensible log entry too.

Issue
https://issues.sonatype.org/browse/NEXUS-8637

Example (newlines added for better readability) log line:
```
2015-05-15 13:30:33,428+0200 INFO  [qtp766265098-51] admin 
org.sonatype.nexus.repository.httpbridge.internal.ViewServlet - 
Illegal operation PUT /repository/site/./css/maven-base.css: Repository 'site' does not allow updating assets.
```

CI
http://bamboo.s/browse/NX3-OSSF312


cstamas:
http://bamboo.s/browse/NX3-OSSF312-1 :+1:

mrprescott:
+1, though I think we soon need to start thinking about factoring out some sort of exception transformer, before this gets too large.

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/StorageTxImpl.java
```diff
@@ -421,14 +421,14 @@ public BlobRef setBlob(final InputStream inputStream, final Map<String, String>
 
     final WritePolicy effectiveWritePolicy = writePolicySelector.select(asset, writePolicy);
     if (effectiveWritePolicy == WritePolicy.DENY) {
-      throw new IllegalOperationException("Repository is read only.");
+      throw new IllegalOperationException("Repository '" + getBucket().repositoryName() + "' is read only.");
```
jdillon: throw new IllegalOperationException("Repository is read only: " + getBucket().repositoryName()) is generally simpler and conforms better to how other exceptions are being created in the system.

jdillon:
@mrprescott I did a while back look at adding exception mappers, but I aborted as it was becoming convoluted and seemed like handlers should be dealing with exceptions to responses.

I agree that we don't want to rebuild similar exception handling mess we had before, which is why I had been living with the noise until it was clear what exceptions there were, if they were even valid to be thrown, etc.

cstamas:
Merged as a06a227b90587b2413fe7dde6cc12172e815ec4b

PR #1322, opened 2015-05-15
[NEXUS-8635] Do not apply opacity to picker labels when disabled

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8635

Before:
roles - sonatype nexus 2015-05-15 14-53-18

After:
roles - sonatype nexus 2015-05-15 14-52-20


adreghiciu:
+1

jdillon:
thanks

PR #1324, opened 2015-05-15
[NEXUS-8630] Stop auto-collapsing the side nav

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8630

Before: http://take.ms/6FIqC

After: http://take.ms/xaGHR


adreghiciu:
+1

PR #1313, opened 2015-05-14
[NEXUS-8549,NEXUS-8518,NEXUS-8609] purge legacy security, add default anonymous roles, remove PluginIdentity

jdillon:
https://issues.sonatype.org/browse/NEXUS-8549
https://issues.sonatype.org/browse/NEXUS-8518
https://issues.sonatype.org/browse/NEXUS-8609

related:

https://github.com/sonatype/nexus-pro/pull/543

includes:

https://github.com/sonatype/nexus-oss/pull/1325

... with modifications to make JS permissions work exactly like Java WildcardPermission

PR #1325, opened 2015-05-15
[NEXUS-8639] Implement wildcard permission logic in UI

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8639


jdillon:
Did some basic tests and this seems to work.  This probably needs to get updated:

```
/**
   * @public
   * @returns {boolean} True if user is authorized for at least one permission that starts with expected string.
   */
  checkAny: function(perm) {
    var me = this,
        hasAny = false;

    if (!me.available()) {
      return false;
    }

    Ext.Object.each(me.permissions, function(key, value) {
      if (Ext.String.startsWith(key, perm) && value === true) {
        hasAny = true;
        return false;
      }
      return true;
    });

    return hasAny;
  }
```

And we probably want to cache the result and invalidate the cache when the permissions change, as this does a lot of work over and over.

jdillon:
checkAny() only used by:

```
visible: function() {
      return NX.Permissions.checkAny('nexus:repository-admin') && NX.State.getUser();
    }
```

Maybe that should use check() instead? and checkAny() can be dropped?

jdillon:
generally +1, needs some additional review of the impliedCache I added and sorting out checkAny()

jdillon:
Does this support checking for nexus:foo:read,write?  There is logic in the ext.direct component to unroll this to nexus:foo:read and nexus:foo:write, which is faulty and generates duplicates.  Would be better if the JS logic behaved the same as the java code, so that may mean some additional work?

adreghiciu:
Closing in favour of https://github.com/sonatype/nexus-oss/pull/1313

PR #1326, opened 2015-05-18
[NEXUS-8606] Remove the LDAP backup mirror feature

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8606

I’d appreciate some eyes on this, as I may have inadvertently broken things. OSS builds and runs fine, and I don’t see any errors when I mess with LDAP connections. Still, a sanity check would be good.

Thanks.


adreghiciu:
There is no point in keeping the FallBackConnectionTest
+1

djsauble:
Manual merge. https://github.com/sonatype/nexus-oss/commit/b56eafb59f445387287035e21caad146049f8563

jdillon:
-1 this should not have been merged.  it a) doesn't build... tests fail, b) leaves ldap feature completely broken

PR #1327, opened 2015-05-18
NEXUS-8305: Explicit DB TX

cstamas:
Explicitly re-using same ODB, and driving TX beginning/end. Preliminary testing shows it works and there is NO need for isolation anymore (as there are no overlapping connections no transactions over single thread).

Issue

CI
http://bamboo.s/browse/NX3-OSSF319


components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/StorageTxImpl.java
```diff
@@ -101,7 +105,9 @@ public StorageTxImpl(final BlobTx blobTx,
     this.assetEntityAdapter = checkNotNull(assetEntityAdapter);
     this.hook = checkNotNull(hook);
 
-    db.begin(TXTYPE.OPTIMISTIC);
+    if (!userManagedDb) {
+      db.begin(TXTYPE.OPTIMISTIC);
```
mrprescott: As discussed on the phone, the transaction-control aspect of StorageTx needs to be made consistent.  The close() and rollback() methods are still dealing with transactions, so the constructor should be calling db.begin() regardless of the user-managed mode.

cstamas: Ack fixed in https://github.com/sonatype/nexus-oss/commit/08b853737084db45647dcf63f5e80219ff8b4813

mrprescott:
+1

cstamas:
Merged as d6a235909f260ec64ab01e98e57dfd797885d95b

PR #1335, opened 2015-05-20
Move content validation to StorageTX, add TX principal name

cstamas:
More improvements to TX, such as:
- move-in content validation (and remove lot of dupe code from plugins)
- pre-fill constant Blob headers (instead doing it in plugins)

Builds upon
https://github.com/sonatype/nexus-oss/pull/1192

CI
http://bamboo.s/browse/NX3-OSSF328


components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/AssetBlob.java
```diff
@@ -61,6 +65,11 @@ void setAttached(final boolean attached) {
   }
 
   @Nonnull
+  public String getBlobName() {
```
mrprescott: Is this used?

mrprescott:
It makes sense to consolidate how content validation gets done. Is there any risk of formats where some content needs validating and some doesn't?

mrprescott:
+1

cstamas:
@mrprescott I too think this is good direction, but will close out this PR and create new one that will factor in format specific validators (based on similar solution as with HttpResponseSender).

PR #1336, opened 2015-05-20
NEXUS-8665: NEXUS-8666: Graceful content validation responses

cstamas:
Changes:
- NX3 should gracefully handle content validation failures
- groups should silently skip members throwing content validation failures

Issues
https://issues.sonatype.org/browse/NEXUS-8665
https://issues.sonatype.org/browse/NEXUS-8666

CI
http://bamboo.s/browse/NX3-OSSF329

### DIFF
--- a/components/nexus-repository/src/main/java/org/sonatype/nexus/repository/group/GroupHandler.java
+++ b/components/nexus-repository/src/main/java/org/sonatype/nexus/repository/group/GroupHandler.java
@@ -21,8 +21,6 @@ import javax.annotation.Nonnull;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
-import org.sonatype.nexus.repository.IllegalOperationException;
-import org.sonatype.nexus.repository.InvalidContentException;
 import org.sonatype.nexus.repository.Repository;
 import org.sonatype.nexus.repository.http.HttpResponses;
 import org.sonatype.nexus.repository.http.HttpStatus;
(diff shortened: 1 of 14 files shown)


mrprescott:
Jason recently mentioned the idea of a handler to do this, which seems like a better idea. If there was a top-of-stack handler in each recipe doing this conversion, then we wouldn't need similar code between ViewServlet and GroupHandler (i.e. to catch IllegalOperationException | InvalidContentException e).  For groups, the original .isSuccessful() check would be enough, since the problem would have been mapped to an http response already.

cstamas:
good idea! will close this one when back

thanks (phone),
~t~

plugins/basic/nexus-repository-httpbridge/src/main/java/org/sonatype/nexus/repository/httpbridge/internal/ViewServlet.java
```diff
@@ -129,6 +131,15 @@ protected void service(final HttpServletRequest httpRequest, final HttpServletRe
       log.info("Illegal operation {} {}: {}", httpRequest.getMethod(), httpRequest.getRequestURI(), e.getMessage());
       send(null, HttpResponses.badRequest(e.getMessage()), httpResponse);
     }
+    catch (InvalidContentException e) {
+      log.info("Invalid content {} {}: {}", httpRequest.getMethod(), httpRequest.getRequestURI(), e.getMessage());
```
jdillon: why logged as info?

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/group/GroupHandler.java
```diff
@@ -142,10 +149,15 @@ protected Response getFirst(final @Nonnull Request request,
...
 
-      final ViewFacet view = member.facet(ViewFacet.class);
-      final Response response = view.dispatch(request);
+      try {
+        final ViewFacet view = member.facet(ViewFacet.class);
+        final Response response = view.dispatch(request);
 
-      responses.put(member, response);
+        responses.put(member, response);
+      } catch (IllegalOperationException | InvalidContentException e) {
```
jdillon: watch your formatting plz

cstamas:
@mrprescott @jdillon updated to use handler

mrprescott:
+1

cstamas:
merged as 805571215aec61ab7cc134f05e72672bfeb8e69f

PR #1358, opened 2015-05-27
Removing redundant logger

mrprescott:
During an earlier PR, I added a logger to NexusPaxExamSupport, but didn't remove the redundant one from NugetITSupport. This corrects that.


cstamas:
+1

PR #1368, opened 2015-05-28
[NEXUS-8644] Hacked fix for misaligned header icon

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8644

Fix involves a call to `setTimeout`, which basically ensures that this isn’t a great solution. Interested in feedback from others. It seems like newly loaded content requires some time to settle before the header layout works correctly. No idea why.

I’ve tried resuming events after `boxready` or `afterrender` fires on the component, but without success. Only timeout seems to work.

None of the drilldown views seem to this problem. Only “Welcome”, “User Accounts”, and “NuGet API Key”, that I’ve seen.


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Content.js
```diff
@@ -104,8 +106,10 @@ Ext.define('NX.controller.Content', {
     // install new feature view
     content.add(cmp);
 
-    // fire activate event to view component
-    cmp.fireEvent('activate', cmp);
```
jdillon: why is this event removed?

jdillon:
seems like there might be need to force a layout thats missing?

But also understanding why this actually broke would be useful, as this only recently broke.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Content.js
djsauble: The purpose of that event is to make the new component visible, but removing it doesn’t seem to break anything. Are components that are added to containers automatically activated? @adreghiciu?

jdillon: @djsauble the purpose is to allow for components to do something after it has been visually activated:

http://docs.sencha.com/extjs/4.2.3/#!/api/Ext.AbstractComponent-event-activate

ie. the view is now visible, give it a chance to do something.

There are a handful of 'activate' event handlers, though I've not looked to see if they are specific to the feature view here or not.

jdillon:
Looks like the css for nx-feature-content icon height/width is incorrect:

```
  $ui-header-icon-height: 32,
  $ui-header-icon-width: 32,
```

should be:

```
  $ui-header-icon-height: 32px,
  $ui-header-icon-width: 32px,
```

After changing nexus-rapture-baseapp/src/main/baseapp/packages/rapture-theme/sass/src/Ext/panel/Panel.scss and regenerating, the welcome page renders properly on the first go.

Unsure why it works on resize or layout change, but its likely because the browser is throwing away the bad css, and instead using the css the icon has on it?

jdillon:
Created https://github.com/sonatype/nexus-oss/pull/1379 to replace this

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Content.js
djsauble: k, thanks for the clarity. I’ll leave it in.

PR #1373, opened 2015-05-29
Repository.getUrl

mrprescott:
Refactoring away from using Response.getRequestUrl to Repository.getUrl.

### DIFF
--- a/components/nexus-repository/src/main/java/org/sonatype/nexus/repository/Repository.java
+++ b/components/nexus-repository/src/main/java/org/sonatype/nexus/repository/Repository.java
@@ -12,9 +12,6 @@
  */
 package org.sonatype.nexus.repository;
 
-import java.net.URI;
-import java.net.URISyntaxException;
-
 import javax.annotation.Nonnull;
 
 import org.sonatype.nexus.repository.config.Configuration;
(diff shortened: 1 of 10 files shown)


components/nexus-repository/src/main/java/org/sonatype/nexus/repository/Repository.java
```diff
@@ -108,4 +111,9 @@
    */
   @Nonnull
   <T extends Facet> T facet(Class<T> type) throws MissingFacetException;
+
+  /**
+  * Returns the URI for the root of the repository, with a final slash.
+   */
+  String getUrl();
```
jdillon: minor mismatch between URL and URI in api/javadoc.

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/Repository.java
```diff
@@ -12,6 +12,9 @@
  */
 package org.sonatype.nexus.repository;
 
+import java.net.URI;
+import java.net.URISyntaxException;
```
jdillon: unused

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/manager/RepositoryImpl.java
```diff
@@ -323,4 +324,9 @@ public void attach(final Facet facet) throws Exception {
 
     return facet;
   }
+
+  @Override
+  public String getUrl() {
+    return BaseUrlHolder.get() + "/repository/" + getName() + "/";
```
jdillon: consider dropping trailing "/" and leave templates below as they were before

jdillon:
+1

PR #1385, opened 2015-06-01
[NEXUS-8694] Add patch for Chrome 43 submenu hover bug

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8694

Note: This is a bug with Chrome 43 (released on May 17th). When Chrome 44 is released, we should reevaluate and remove the patch, if fixed.


jdillon:
+1

mrprescott:
+1

PR #1405, opened 2015-06-08
[NEXUS-8587] Change the naming convention of I18N strings

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8587

Related: https://github.com/sonatype/nexus-pro/pull/560

Lots of manual changes here. I ran i18nverifier to make sure everything lines up, but I undoubtedly fat-fingered a few string names somewhere.


adreghiciu:
+1

kellyrob99:
I realize that I am late to the party here and wasn't around to debate when we planned it, but I do not see this change as beneficial. Strings are now longer, do not appear as constants due to the change in case, and appear completely bound to path information - which inevitably will complicate any future refactorings. Personally I would prefer to keep the UPPERCASE convention for what are effectively constants, and in all cases aim for shortening the keys for clarity rather than expanding them for specificity. I will defer to the team, but I am -1 on this particular change.

kellyrob99:
And just to confirm, this would not be included in M4 release in any case, correct?

djsauble:
Correct. This change will not be included in the M4 release.

I’ve copied you on the original discussion thread.

mosabua:
The only concern I would have is that if the same user facing string is used in two places you need two separate i18n keys and entries. E.g. say use have a string that is just "Component" .. and you have to use that in several places in the UI. Imho it should be defined in ONE property and not in multiple and then have to be ensured to be consistent.

mosabua:
In terms of upper vs camel case I have used both in apps in the past and both have their pros and cons. More a question of taste so whatever you decide as a team ;-)

djsauble:
@mosabua Things like “Submit” and “Cancel” buttons should certainly have their strings reused. I’ve chosen not to do that for now, since it’s more complicated to manage 1:N mappings, and docs should be involved with that, to help us decide which strings to reuse (or not).

But yes, we should do that eventually. It’s just a bigger effort I don’t want to tackle right now.

jdillon:
@mosabua @djsauble yes we certainly could setup some global strings for common things that will only be used in a single context, like the text of common buttons.  But I don't think its worth trying to optimize out the duplication for everything, as there are and will continue to be times when we do want some flexibility to customize those as needed.  And if we reduce all the duplicates, we'd have to then re-extract new keys for the places we want that changed and I can see that becoming a bit of a mess.

@mosabua the point of our use of NX.I18n is not to reduce string duplication.  It is to allow a single (or small set) of places to go and see the user-facing text and change it.  So that we do decide to change the verbiage that its a much simpler task focused on a 1-3 js files.

jdillon:
@djsauble I suggest we drop the starting prefixes, like 'Coreui_Controller_' and 'Coreui_View_'.  There is a small chance that plugins will collide, but I don't think that is presently going to happen.  Point out any concerns you have here though.  FeatureGroups is one that comes to mind, but the strings it defines shouldn't overlap.

We can leave the other simplifications due to package-suffix and classname-prefix to later, but probably can do that eventually and simplify the strings even more.

jdillon:
@djsauble @kellyrob99 also about UPPERCASE vs. CamelCase I do feel like the later is easier to read, so I suggest we retain this.  @kellyrob99 as mentioned on call, don't think of them as constants (in java-like terms), think of them message-bundle keys.

djsauble:
Cool, thanks for the feedback. Next steps for me:

+ Eliminate prefixes (e.g. `Coreui_Controller_`)
+ Remove multiple instances of the same word in a key (e.g. `Analytics_AnalyticsSettings` becomes `Analytics_Settings`)

@kellyrob99 Correct me if wrong, but I believe the above addresses your concerns about strings being tied to absolute paths. Technically, the name is still derived from the path, but it’s simplified enough to be more descriptive than prescriptive. Agree/disagree?

kellyrob99:
@djsauble Yes, this should make things less verbose and a little less brittle - works for me!

jdillon:
@djsauble I would avoid unrolling the Analytics_AnalyticsSettings to Analytics_Settings stuff until we fix the underlying classnames. otherwise +1

jdillon:
+1 generally looks better

adreghiciu:
+1 After the latest master merge there are strings from browse UI that are not using the latest convention

djsauble:
@adreghiciu Fixed.

djsauble:
Manual merge. https://github.com/sonatype/nexus-oss/commit/c729915134b1cf0a8b04a37f167cc1638b3fd88a

PR #1407, opened 2015-06-08
NEXUS-8718: Use standard HTTP proxy host exclusion patterns

cstamas:
Drop use of regexp, instead use Java standard glob-like expressions. Still, as _there is no "standard ASF HttpClient implementation"_ (HttpClient uses `ProxySelector.getDefault()` that is backed by `sun.net.spi.DefaultProxySelector`) some simple trick is applied for hostname matching. Apply validator and change UI label. Also, this PR will align NX non-proxy host behaviour with Java, by appending the default exclusions (as seen in sun.net.spi.DefaultProxySelector).

Post M4

Issue
https://issues.sonatype.org/browse/NEXUS-8718

CI
http://bamboo.s/browse/NX3-OSSF403


jdillon:
Still confused, the JDK impl of this should do the globing already, why we do have to dup it?

cstamas:
@jdillon because the JDK implementation DefaultProxySelector sources those directly from system properties, and the underlying actual matching implementation is `sun.misc.RegexpPool` that is private.

cstamas:
23eb9f168ebbe5f5aaeeced28dcd56ce570246d5

PR #1409, opened 2015-06-08
Support encoded urls in base url detection.

bradbeck:
https://issues.sonatype.org/browse/NEXUS-8791


components/nexus-core/src/test/java/org/sonatype/nexus/internal/app/BaseUrlManagerImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+  Provider<HttpServletRequest> requestProvider;
+
+  @Mock
+  HttpServletRequest request;
+
+  BaseUrlManagerImpl underTest;
+
+  @Test
+  public void forceUrl() {
+    underTest = new BaseUrlManagerImpl(requestProvider);
```
jdillon: this seems the same for every-test, why not do this in @Before setUp() and avoid duplication?

components/nexus-core/src/test/java/org/sonatype/nexus/internal/app/BaseUrlManagerImplTest.java
```diff
@@ -0,0 +1,89 @@
...
+  Provider<HttpServletRequest> requestProvider;
+
+  @Mock
+  HttpServletRequest request;
+
+  BaseUrlManagerImpl underTest;
+
+  @Before
+  public void setUp() {
+    underTest = new BaseUrlManagerImpl(requestProvider);
```
kellyrob99: You should be able to omit even this, if you annotate underTest with @InjectMocks it should be instantiated for you.

bradbeck: Good point. Thanks!

components/nexus-core/src/test/java/org/sonatype/nexus/internal/app/BaseUrlManagerImplTest.java
```diff
@@ -0,0 +1,85 @@
...
+  static final String FAKE_URL = "http://example.com:1234/foo/bar";
+
+  @Mock
+  Provider<HttpServletRequest> requestProvider;
+
+  @Mock
+  HttpServletRequest request;
+
+  @InjectMocks
+  BaseUrlManagerImpl underTest;
```
jdillon: ugh, prefer to avoid all things "inject" over simple @Before usage to setup the component under test.

kellyrob99: why? we're avoiding framework injection and utilizing a built-in feature of mockito so I don't understand your objection here

jdillon:
overall +1 but want to ensure we do some testing to ensure this does work when context changes, with reverse-proxy, etc.

this is the right direction, didn't read the javadoc closely before to realize that some where encoded and some were not.

Unsure about double encoding too, but since this only needs to find the base-url, this is likely limited just to the servlet-api and if we trust its returning the right things and using the right api, then it should be fine.

bradbeck:
Replaced by: https://github.com/sonatype/nexus-oss/pull/1422

PR #1411, opened 2015-06-09
[NEXUS-8783] Add isPermittedAny() and isPermittedAll() conditions

jdillon:
https://issues.sonatype.org/browse/NEXUS-8783

If this is sane, there are a bunch of other places to apply isPermittedAny to.

Also added Conditions.isPermittedAll()


adreghiciu:
-1 https://github.com/sonatype/nexus-oss/pull/1412

kellyrob99:
Seems sane, but reserving my blessing until @adreghiciu 's concerns are answered and his suggested replacement PRs evaluated

jdillon:
@adreghiciu cool, glad its not as wide spread as it looked yesterday ;-)

documentation around these could have avoided some confusion, ie. documenting the permissions on the Form is for edit, only?  Or define the permissions in the wrapper around the form?

jdillon:
will drop, may retain the helpers, though both could be done with Condition logic, so not strictly needed, though IsPermittedAny slightly useful to avoid needing a nested OR condition.

jdillon:
dropping in favor of other changes

PR #1413, opened 2015-06-09
Fix editable condition/marker for capabilities add/update

adreghiciu:


plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/capability/CapabilitySettingsForm.js
```diff
@@ -50,6 +50,18 @@ Ext.define('NX.coreui.view.capability.CapabilitySettingsForm', {
 
   /**
    * @override
+   */
+  initComponent: function() {
+    var me = this;
+
+    me.editableCondition = me.editableCondition || NX.Conditions.isPermitted('nexus:capabilities:update');
```
djsauble: Is the default condition ever overridden? If not, why the `||`?

adreghiciu: It is overridden in CapabilityAdd

djsauble: k, +1 from me

kellyrob99:
+1, provided #1411 doesn't conflict

PR #1421, opened 2015-06-10
POC of direct use of OrientDB cache clear in retries

cstamas:
This is just a POC PR, that simplifies the retry handling, allowing even `StorageTx` to be passed in (basically solving all the ugliness of previous workarounds). Also adds IOEx.

Cons:
* StorageFacetImpl uses directly `tx.getDb().commit`. This clearly separates the TX into "orient TX" and "the rest TX" (blobs). But it may be actually a signal that we need something like this? https://gist.github.com/cstamas/5f5db1be131847b2458a


mrprescott:
I'm assuming that Orient's tx isolation and the cache are separate things; if so, fiddling with the cache to improve how transactions play together seems like the wrong way to fix the problem. If we (for whatever reason) moved to 'repeatable read' isolation, cache-clearing would do nothing, wouldn't it? It only seems to do something now because we're using 'read committed' tx isolation.

cstamas:
Am more and more inclined to say that "retries" (the recommended approach if you hit concurrency issue with Orient DB updates/changes) are actually not "improving how transactions play together", as we still enforce single transactions only (no nested ones allowed). They are more a possible way to cope with repeated performs. I see them basically as SQL `NOCACHE` keyword is, but "globally applied" :smile:

cstamas:
Closing this PR, as it was just a POC, example of idea to look at.

PR #1423, opened 2015-06-10
[NEXUS-8772] Use a setter instead of directly update private properties

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8772


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/util/condition/IsPermitted.js
```diff
@@ -50,6 +50,20 @@ Ext.define('NX.util.condition.IsPermitted', {
...
+   * @public
+   *
+   * Sets permission and re-evaluate.
+   *
+   * @param {String} permission permission
+   */
+  setPermission: function(permission) {
+    var me = this;
+
+    me.permission = permission;
```
jdillon: is me.permission meant to be private then?  perhaps mark it as so then

jdillon:
+1

kellyrob99:
+1

PR #1428, opened 2015-06-11
NEXUS-8725: Privilege and Role descriptions

cstamas:
https://issues.sonatype.org/browse/NEXUS-8725

CI
http://bamboo.s/browse/NX3-OSSF423


jdillon:
+1 looks reasonable

plugins/security/nexus-ldap-plugin/src/main/java/org/sonatype/nexus/ldap/internal/StaticSecurityConfigurationResourceImpl.groovy
```diff
@@ -67,6 +71,7 @@ class StaticSecurityConfigurationResourceImpl
             ),
             new CPrivilege(
                 id: 'ldap-delete',
+                description: 'Delete permission for Ldap',
```
joedragons: As an acronym should these be "LDAP"?  I am not sure if that's consistant across the site but I feel like it is mostly.

adreghiciu:
+1 What about description of dynamically generated permissions?

cstamas:
413e2ab48069af0ebab2d4caa7df1983cc667226

jdillon:
ugh

-1 the dynamic permission descriptions added are not proper

PR #1430, opened 2015-06-11
NEXUS-8821 - improve logging for failed auth token request

peterlynch:
make it easier to diagnose username mismatch errors

https://issues.sonatype.org/browse/NEXUS-8821

### DIFF
--- a/plugins/internal/nexus-wonderland-plugin/src/main/java/org/sonatype/nexus/wonderland/rest/AuthenticateResource.java
+++ b/plugins/internal/nexus-wonderland-plugin/src/main/java/org/sonatype/nexus/wonderland/rest/AuthenticateResource.java
@@ -91,11 +91,15 @@ public class AuthenticateResource
     final String password = Tokens.decodeBase64String(token.getP());
 
     // Require current user to be the requested user to authenticate
-    Subject subject = security.getSubject();
-    final String principalName =  subject.getPrincipal().toString();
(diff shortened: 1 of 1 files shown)


plugins/internal/nexus-wonderland-plugin/src/main/java/org/sonatype/nexus/wonderland/rest/AuthenticateResource.java
```diff
@@ -87,13 +87,17 @@ public AuthenticateResource(final SecuritySystem security,
...
+    final String password = Tokens.decodeBase64String(token.getP());
 
     // Require current user to be the requested user to authenticate
     Subject subject = security.getSubject();
-    if (!subject.getPrincipal().toString().equals(username)) {
+    final String principalName =  subject.getPrincipal().toString();
+    log.debug("payload username: {}, payload password: {}, principal: {}", username, Tokens.mask(password), principalName);
+
+    if (!principalName.equals(username)) {
+      log.warn("access token request denied - authenticated user {} does not match payload user {}",
```
jdillon: auth-token not access token probably clearer to devs.

jdillon:
+1 after the string is updated

cstamas:
+1

adreghiciu:
+1

jdillon:
should update master as well, it still has this endpoint so we should update it as well.

peterlynch:
I will make a master PR when Rich is back

PR #1435, opened 2015-06-12
NEXUS-8791 integration test for URL encoded ODATA package queries

bradbeck:
An integration test to ensure that the link returned in the package metadata contains a usable URL to the package.

https://issues.sonatype.org/browse/NEXUS-8791

### DIFF
--- a/testsuite/modern-testsuite/src/test/java/org/sonatype/nexus/testsuite/nuget/NugetHostedIT.java
+++ b/testsuite/modern-testsuite/src/test/java/org/sonatype/nexus/testsuite/nuget/NugetHostedIT.java
@@ -12,6 +12,7 @@
  */
 package org.sonatype.nexus.testsuite.nuget;
 
+import java.net.URLEncoder;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


mrprescott:
+1

testsuite/modern-testsuite/src/test/java/org/sonatype/nexus/testsuite/nuget/NugetHostedIT.java
```diff
@@ -175,4 +176,23 @@ public void visualStudioSearchForUpdates() throws Exception {
     assertThat("entry count", entries.size(), is(1));
     assertThat("entry ID", entries.get(0).get("ID"), is("SONATYPE.TEST"));
   }
+
+  /**
+   * Ensure that an ODATA query for an uploaded package's metadata
+   * returns an entry that contains a valid content link to the package.
+   */
+  @Test
+  public void link_to_package_is_valid_in_package_metatdata_returned_from_url_encoded_ODATA_query() throws Exception {
```
kellyrob99: Would prefer this to be a shorter camelcase name and depend on the jdoc for detail. Can't say this style is actually against our coding guidelines, but it would certainly be a departure from the norm.

cstamas:
+1

bradbeck:
http://bamboo.s/browse/NX3-OSSF432-1

PR #1439, opened 2015-06-15
Conditional HTTP requests

cstamas:
Support for 4 basic HTTP conditional headers (If-Modified-Since, If-Unmodified-Since, If-Match, If-None-Match).

CI
http://bamboo.s/browse/NX3-OSSF430


components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/ConditionalRequestHandler.java
```diff
@@ -0,0 +1,85 @@
...
+    final Predicate<Response> requestPredicate = requestPredicate(context.getRequest());
+    if (requestPredicate != null) {
+      final String action = context.getRequest().getAction();
+      switch (action) {
+        case GET:
+        case HEAD: {
+          final Response response = context.proceed();
+          if (response.getStatus().isSuccessful()) {
+            if (!requestPredicate.apply(response)) {
+              return new Response.Builder().status(Status.failure(NOT_MODIFIED, "Not modified")).build();
```
kellyrob99: This shouldn't be marked as a failure, should it?

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/ConditionalRequestHandler.java
```diff
@@ -0,0 +1,85 @@
...
+          }
+          return response;
+        }
+
+        case POST:
+        case PUT:
+        case DELETE: {
+          final Response getResponse = performGet(context.getRepository(), context.getRequest());
+          if (getResponse.getStatus().isSuccessful() && !requestPredicate.apply(getResponse)) {
+            return new Response.Builder().status(Status.failure(PRECONDITION_FAILED, "Precondition failed")).build();
```
kellyrob99: Message here is not terribly helpful to clients, can we improve?

kellyrob99:
Needs at least unit tests please ;)

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/ConditionalRequestHandler.java
cstamas: Status message set is actually the "reason phrase" as per RFC http://tools.ietf.org/html/rfc7230#section-3.1.2

And should not be used as clients are advised to ignore it completely.

OTOH, added toString method to HTTP conditions, so doing a toString() on returned non-null `requestPredicate` should describe just nicely what conditions are (see added DEBUG log too)

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/ConditionalRequestHandler.java
```diff
@@ -0,0 +1,99 @@
...
+        requestPredicate);
+    switch (action) {
+      case GET:
+      case HEAD: {
+        final Response response = context.proceed();
+        // keep all response headers like Last-Modified and ETag, etc
+        if (response.getStatus().isSuccessful() && !requestPredicate.apply(response)) {
+          return new Response.Builder()
+              .copy(response)
+              .status(Status.failure(NOT_MODIFIED, "Not modified"))
```
jdillon: I also don't see the point in duplicating this, does the underlying system not provide the default reason-phrase?  If not perhaps that is the right change to make instead of this duplication of meaningless strings.

jdillon: FTR the RFC you linked, instructs clients to ignore this value.

cstamas:
5d41322463df3d2a31bfa1efe5f8c3aea353e086

PR #1440, opened 2015-06-15
[NEXUS-8810] Allow repository to select blobstore

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8810


components/nexus-blobstore-api/src/main/java/org/sonatype/nexus/blobstore/api/BlobStore.java
```diff
@@ -120,4 +120,19 @@
...
    */
   AutoClosableIterable<BlobId> iterator();
+
+  /**
+   * @return true if this blob store is is in use
+   */
+  boolean isInUse();
+
+  /**
+   * Increases the usage of this blob store;
```
mrprescott: I thought this was some instruction to use the blob store more, or less. Perhaps:
"Increments this blob store's the usage counter"

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/controller/Blobstores.js
```diff
@@ -155,6 +155,27 @@ Ext.define('NX.coreui.controller.Blobstores', {
...
+   * @protected
+   * Enable 'Delete' when user has 'delete' permission, blob store is not in use and it does not have any blob.
+   */
+  bindDeleteButton: function(button) {
+    var me = this;
+    button.mon(
+        NX.Conditions.and(
+            NX.Conditions.isPermitted(me.permission + ':delete'),
+            NX.Conditions.gridHasSelection(me.masters[0], function(model) {
+              return !model.get('inUse') && model.get('blobCount') === 0;
```
mrprescott: Do we need a corresponding server-side check, so a stale UI can't delete a blob store if another user has just associated it with a repository?

components/nexus-blobstore-api/src/main/java/org/sonatype/nexus/blobstore/api/BlobStore.java
adreghiciu: Renamed

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/controller/Blobstores.js
adreghiciu: Added a check to blob store manager when deleting

mrprescott:
+1

cstamas:
+1

jdillon:
could have sworn we were going to allow this to be set on create only and defer larger change to cope with deletion until it was thought out more?

mrprescott:
@jdillon Quite right, we did, it's in the issue's AC.

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/app/PluginStrings.js
```diff
@@ -232,6 +232,9 @@ Ext.define('NX.coreui.app.PluginStrings', {
     ADMIN_REPOSITORIES_SETTINGS_CONNECTION_RETRIES_HELP: 'How many times to try to connect before giving up',
     ADMIN_REPOSITORIES_SETTINGS_CONNECTION_TIMEOUT: 'Connection timeout',
     ADMIN_REPOSITORIES_SETTINGS_CONNECTION_TIMEOUT_HELP: 'Seconds to wait for activity before stopping and retrying the connection. Leave blank to use the globally defined HTTP timeout.',
+    ADMIN_REPOSITORIES_SETTINGS_BLOBSTORE: 'Blobstore',
+    ADMIN_REPOSITORIES_SETTINGS_BLOBSTORE_HELP: 'Blobstore to be used to store the assets.',
```
jdillon: maybe 'Blobstore to be used to store asset contents' would be better

jdillon:
@mrprescott then the incr/decr stuff seems a bit premature, could be the right thing, but i'm not sure if we have considered how blobstores will interact with other nodes in the near future to build that yet.

adreghiciu:
Reverted blobstore delete conditions/usage count

adreghiciu:
Manually merged https://github.com/sonatype/nexus-oss/commit/51c0f6dd9d539f5283927232c2cc8ddc6c971a8c

PR #1443, opened 2015-06-16
Maven default content attributes

cstamas:
Last-Modified was recorded even now, but ETag was not. It was
only stored IF remote proxy did send it (usually did), but in case
of hosted repositories, ETag was never present, as client/wagon
never sends it.


plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/HostedHandler.java
```diff
@@ -90,4 +97,20 @@ private Response doDelete(final MavenPath path, final MavenFacet mavenFacet) thr
     }
     return HttpResponses.noContent();
   }
+
+  /**
+   * Adds {@link HttpHeaders#ETAG} header if not present. In case of hosted repositories, this is safe and even
+   * good thing to do, as the content is hosted here only.
+   */
+  private void mayFormatETag(final Content content) {
```
mrprescott: Minor: Perhaps this should be 'mayAddETag'.

mrprescott:
+1

plugins/basic/nexus-repository-maven/src/main/java/org/sonatype/nexus/repository/maven/internal/MavenFacetImpl.java
```diff
@@ -375,9 +376,15 @@ private void putAssetPayload(final StorageTx tx,
...
+      final String etag = contentAttributes.get(Content.CONTENT_ETAG, String.class);
+      if (!Strings.isNullOrEmpty(etag)) {
+        formatAttributes.set(P_CONTENT_ETAG, etag);
+      }
+      else {
+        formatAttributes.remove(P_CONTENT_ETAG);
+      }
     }
-    else {
+    if (formatAttributes.get(P_CONTENT_LAST_MODIFIED) == null) {
```
kellyrob99: this might make more sense as an else block for the lastModified != null check above rather than spread the logic out

kellyrob99: Never mind - I see that the other check depends on nullity of the overall contentAttributes

kellyrob99:
+1

cstamas:
131a004177b087377f386c70192888a4c2f143af

PR #1450, opened 2015-06-17
NEXUS-8811: Blob store addressing

cstamas:
https://issues.sonatype.org/browse/NEXUS-8811

http://bamboo.s/browse/NX3-OSSF447


mrprescott:
This seems to address the write side, but presumably the read side is a little more involved as a BlobTx could be dealing with blobs spanning multiple blob stores.

jdillon:
@mrprescott yes, we'll also need to plan for remote blobs too, so its okay to have a few baby steps here to get to the desired end state.

mrprescott:
@jdillon Good - after quick chat with Tamas, I realized my comment taken at face value would unnecessarily blow the scope of this out. Given that Alin's work will prevent repositories changing blob stores, we don't need to add any code now to plan for blobs in multiple blob stores. Perhaps just a comment to note that at some point, the assumption that a repository's blobrefs can be handled by its one and only current blob store is simplistic and will probably change.

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/storage/StorageFacetImpl.java
```diff
@@ -280,7 +285,7 @@ private StorageTx openStorageTx(final ODatabaseDocumentTx db, final boolean isUs
     return StateGuardAspect.around(
         new StorageTxImpl(
             createdBy(),
-            new BlobTx(blobStore),
+            new BlobTx(localNodeAccess.getId(), blobStore),
```
jdillon: maybe slightly saner to pass in the localNodeAcess and fetch the id when needed closer to the blobref creation.

cstamas:
bf1dab0adaac5ed6240e28bd2519d70f026f0bec

PR #1455, opened 2015-06-18
NEXUS-8850 add uptime log message on shutdown

kellyrob99:
- simple log message using jodatime for formatting

https://issues.sonatype.org/browse/NEXUS-8850


jdillon:
change should probably be targeted at nexus2

nexus3 probably should do this in bootstrap, as there is still some time between stop of NxApplication and the server still running.

jdillon:
FTR I have no objection to merging this asis for now.

PR #1465, opened 2015-06-22
NEXUS-8890 on refresh load Task store too

bradbeck:


plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/controller/Tasks.js
```diff
@@ -251,14 +251,14 @@ Ext.define('NX.coreui.controller.Tasks', {
...
-   * (Re)load task type store && reset all cached combo stores.
+   * (Re)load task store.
    */
   onRefresh: function() {
     var me = this,
         list = me.getList();
 
     if (list) {
-      me.getStore('TaskType').load();
+      me.getStore('Task').load();
```
jdillon: suppose its possible both stores could need to be reloaded?  though TaskType probably won't change too often, but if you were to disable a plugin, it might... /me shrugs

jdillon: FTR I'd +1 this if we reload both stores.

djsauble:
+1

PR #1468, opened 2015-06-23
Improved context search request

mosabua:
replace _ with spaces in link to search to produce nicer results and output there.

https://issues.sonatype.org/browse/NEXUS-8887

This changes a search request for a feature from e.g. 
https://books.sonatype.com/nexus-book/3.0/reference/search.html?q=admin_security_privileges
to
https://books.sonatype.com/nexus-book/3.0/reference/search.html?q=admin%20security%20privileges

(%20 is URLencoded space)

This removes the message visible in 

screen shot 2015-06-23 at 3 15 39 pm

changing it to the nicer

screen shot 2015-06-23 at 3 16 45 pm


mosabua:
Could have also changed this in the controller at 

https://github.com/sonatype/nexus-oss/blob/master/plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Features.js#L95

but was not sure if this would be more desired from your perspective. Could also have used " " instead of "%20" potentially. Same idea.. whatever desired..

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Help.js
```diff
@@ -142,7 +142,7 @@ Ext.define('NX.controller.Help', {
    */
   onFeatureHelp: function() {
     var me = this,
-        keyword = me.selectedFeature.get('helpKeyword'),
+        keyword = me.selectedFeature.get('helpKeyword').replace(new RegExp('_', 'g'), '%20'),
```
mosabua: This is bascially a replaceAll .. wasnt sure if you want to use some sort of function you already got so I just made it work in straight JS. Could also add some sort of comment if you want that in the JS files.

jdillon:
doing this when defining the features, probably changing to helpKeywords is better than hacking this in like this.

mosabua:
@jdillon if I set this in the feature should I use " " instead of "_" or "%20"  (already URL encoded basically) and also .. I assume changing helpKeywords there has no surprising side effects

Testing that now btw..

mosabua:
So spaces and the feature version works. I will append this PR with that solution as @jdillon prefers that. Which makes this a PR a one character change ... hahaha

mosabua:
Done. Kicked off feature build at http://bamboo.s/browse/NX3-OSSF467-1

kellyrob99:
+1

bradbeck:
+1

mosabua:
Manually merged with squash. See https://github.com/sonatype/nexus-oss/commit/07852b9f898e64a20b219273610580658ad4b8a5

PR #1478, opened 2015-06-27
UI logging refinement

jdillon:
Adds logging panel to developer tools.

Introduces NX.Console, replacement for what NX.Log previously did.  And uses NX.Log as a bridge adapter to NX.Console for boot, and then to the NX.controller.Logging once bootstrap has finished.

Adds remoting of log-events (hacky ATM) to server.


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/app/Application.js
```diff
@@ -238,8 +238,8 @@ Ext.define('NX.app.Application', {
 
     remotingProvider = Ext.direct.Manager.addProvider(NX.direct.api.REMOTING_API);
 
-    // disable buffering of method calls
-    remotingProvider.enableBuffer = false;
+    // configure Ext.Direct buffer-window milliseconds
+    remotingProvider.enableBuffer = 100;
```
kellyrob99: Is this going to have any conflict with existing code? It was set to false as part of browse-ui, although I'm not sure why it was necessary: https://github.com/sonatype/nexus-oss/commit/ed97ec82bb17190b36ad45aed67604a485628a78

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Logging.js
```diff
@@ -0,0 +1,141 @@
...
+      event.timestamp = new Date();
+    }
+
+    me.getStore('LogEvent').add(event);
+
+    // HACK: experimental: remote events to server
+    if (me.remote) {
+      // HACK: kill timestamp... GSON freaks out
+      var copy = Ext.clone(event);
+      delete copy.timestamp;
```
kellyrob99: in other places we just use the long value, so Date.now() should work. Alternatively we should be able to register a serializer/de-serializer based on Date string value http://stackoverflow.com/questions/6873020/gson-date-format

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/store/LogEvent.js
```diff
@@ -0,0 +1,23 @@
...
+/*global Ext*/
+
+/**
+ * LogEvent store.
+ *
+ * @since 3.0
+ */
+Ext.define('NX.store.LogEvent', {
+  extend: 'Ext.data.Store',
+  model: 'NX.model.LogEvent'
```
kellyrob99: sorting in descending order on the timestamp here would make sense(most recent at the top) and trivially resolve the TODO you added to NX/controller/dev/Logging.js for moving to the end of the list

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Logging.js
```diff
@@ -0,0 +1,141 @@
...
+  setThreshold: function (threshold) {
+    this.threshold = threshold;
+  },
+
+  /**
+   * Mapping of NX.store.LogLevel weights.
+   *
+   * @private
+   */
+  levelWeights: {
```
kellyrob99: this might be better encoded into the LogLevel model/store to prevent duplicating that information

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/dev/Logging.js
```diff
@@ -0,0 +1,72 @@
...
+  emptyText: 'No events',
+  viewConfig: {
+    deferEmptyText: false
+  },
+
+  columns: [
+    { text: 'level', dataIndex: 'level' },
+    { text: 'logger', dataIndex: 'logger', flex: 1 },
+    { text: 'message', dataIndex: 'message', flex: 3 },
+    { text: 'timestamp', dataIndex: 'timestamp', width: 300 }
```
kellyrob99: why fixed width instead of flex here? Need to be cognizant that fixed widths will be problematic across screen sizes

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/dev/Stores.js
```diff
@@ -33,56 +35,64 @@ Ext.define('NX.view.dev.Stores', {
     var me = this;
 
     Ext.apply(me, {
```
kellyrob99: not directly pertinent to this review, but isn't this 'apply' call completely redundant? Just declaring me.items and me.tbar within the initComponent block should be sufficient. I only mention because we have a mix of styles and picking one consistently would aid comprehension.

kellyrob99:
Seems like a solid refactor and I will work on adding logging to the UI with future dev.
+1

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Logging.js
jdillon: I thought we already had this handling for Dates but I guess we didn't do it, or it got removed or something?

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Logging.js
jdillon: I actually did that, then reverted, as the lookup of store, resolving record, then resolving fields was much more complex that this simple lookup.  So I undid that and went with this simple mapping.  I had considered priming this when the store loads, but the risk of that static store getting out of sync is very low, since we aren't likely to be adding more levels.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/dev/Logging.js
jdillon: because the data is fixed, and if the fonts are in the ballpark, should be close enough.  not overly worried about dev-panel layouts though.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/dev/Stores.js
jdillon: /me shrugs, its a mixed bag IMO, depends on if you prefer foo: bar or me.foo=bar style mostly.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Logging.js
kellyrob99: Did not confirm, but it's likely just that the formatted Date string isn't compatible with what we have declared on the java side: https://github.com/sonatype/nexus-oss/blob/master/plugins/rapture/nexus-extdirect-plugin/src/main/java/org/sonatype/nexus/extdirect/internal/ExtDirectGsonBuilderConfigurator.java#L36-36

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/store/LogEvent.js
jdillon: agree, though I'd still like to learn how to properly scroll, we kinda do that for the logging viewer...

jdillon:
manually merged

PR #1479, opened 2015-06-29
Align Maven facet attribute names with the rest of NX3

cstamas:
Changes:
- align ODB Maven properties to NX3 standard (using_underscore naming instead of "camelCase")
- fix dangling/string literal uses of column names in MD rebuilder

http://bamboo.s/browse/NX3-OSSF482


adreghiciu:
+1

jdillon:
why?  these are format-specific attributes, why not leave the format specific naming style.  in maven its artifactId not artifact_id.

cstamas:
@jdillon I vaguely remember we agreed to align attributes in DB with those in UI, and stick with `underscore_case`, but I might be wrong...

mrprescott:
My recollection was that we agreed to use underscores rather than camel case throughout. shrug

jdillon:
@cstamas @mrprescott for most entities yes, we did that because it will normalize the case of those, so fooBar -> FOOBAR and thats not as easy to read as foo_bar -> FOO_BAR.

But I believe for format-specific attributes, where possible, we use the keys the format knows.  This will avoid us having to add more translation from the format specifics to our specifics.  And in cases this leaks up to the UI, so search would have to change for this, etc.

If you want to use foo_bar style for other non-format stuff go ahead, but I believe for the format specific attributes, use the formats names.

cstamas:
@jdillon ack, will return property names to their original. Still, this PR contains other good things, like fixes for string literal uses of properties (in MD rebuilder), and introduction of MavenAttributes that exposes properties for reuse across plugin.

cstamas:
http://bamboo.s/browse/NX3-OSSF482-4 (with FT failures)

cstamas:
Reverted the attribute names back to "normal" as they were before, merged as 69df4bb45260948f4520943841792caf438d161d

PR #1480, opened 2015-06-29
Format neutral content headers handler

cstamas:
As original MavenHeadersHandler was not maven specific at all,
it adds usual HTTP response headers that is/should be common
for all.

http://bamboo.s/browse/NX3-OSSF483


adreghiciu:
+1

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/ContentHeadersHandler.java
```diff
@@ -10,18 +10,12 @@
  * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
  * Eclipse Foundation. All other trademarks are the property of their respective owners.
  */
-package org.sonatype.nexus.repository.maven.internal;
+package org.sonatype.nexus.repository.view;
```
jdillon: There is a view.handlers package where standard handles are meant to live, why aren't you using that?

kellyrob99:
+1

components/nexus-repository/src/main/java/org/sonatype/nexus/repository/view/ContentHeadersHandler.java
cstamas: Uh oh, agreed, I just noticed there are a wealth of other handlers in this package too, will move 'em here too.

cstamas:
Merged as 9eba276e7e52fbe1d5a7bf2765f7595133c63110, will move handler to their place directly on master

cstamas:
0e1a0ea7c5a6f5de8034473a91e2c1c4023ac5ac commit moves all the handler to proper place.

PR #1482, opened 2015-06-29
NEXUS-8914: Indicate verified on first GET

cstamas:
And not only on 304 response

Detected by:
https://github.com/sonatype/nexus-oss/pull/1481

Issue
https://issues.sonatype.org/browse/NEXUS-8914

CI
Tbd


components/nexus-repository/src/main/java/org/sonatype/nexus/repository/proxy/ProxyFacetSupport.java
```diff
@@ -215,6 +215,7 @@ protected Content fetch(String url, Context context, Content stale) throws IOExc
       result.getAttributes().set(Content.CONTENT_LAST_MODIFIED,
           extractLastModified(request, response.getLastHeader(HttpHeaders.LAST_MODIFIED)));
       result.getAttributes().set(Content.CONTENT_ETAG, extractETag(response.getLastHeader(HttpHeaders.ETAG)));
+      indicateVerified(context);
```
mrprescott: This is premature, isn't it? At this point, the Content hasn't been stored.

In any case, I think this logic should be handled by the format-specific store() method, to avoid a separate Orient transaction.

cstamas: Ack, as per phone, this is wrong solution. But, FTR, this _does fix_ the problem.

cstamas:
Closing out this one for now, looking for proper solution

PR #1486, opened 2015-06-29
Missing Ext.clone causing memory leaks with drilldown

djsauble:
Multiple components with references to the same object prevented that object from being destroyed and removed from the DOM.

Fixes: http://screencast.com/t/km3hvxPpB5C


djsauble:
Superseded by https://github.com/sonatype/nexus-oss/pull/1486

PR #1494, opened 2015-06-30
[NEXUS-8814] Index / Search Smoke functional tests

adreghiciu:
https://issues.sonatype.org/browse/NEXUS-8814
http://bamboo.s/browse/NX3-OSSF493


testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/harness.js
```diff
@@ -57,6 +57,19 @@ NX.TestHarness.start(
     },
 
     {
+      group: 'Search',
+      items: [
+        {
+          group: 'FT',
+          items: [
+            { url: 'testsuite-oss/search/search_basic.t.js' }
```
kellyrob99: this test will never pass in the browser due to the dependency on Java code, correct?

adreghiciu: Yes

kellyrob99: In that case I wouldn't even bother adding it here, just leave it for the maven builds to pick up

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/search/search_basic.t.js
```diff
@@ -0,0 +1,152 @@
...
+    waitFor: 'storesToLoad', args: function() {
+      return t.cq1('nx-coreui-search-result-list').getStore();
+    }
+  };
+
+  t.describe('Basic Indexing and Searching', function(t) {
+    t.it('Keyword search finds maven components', function(t) {
+      t.chain(
+          t.openPageAsAdmin('browse/search'),
+          function(next) {
```
kellyrob99: openPageAsAdmin() call now includes a waitForAnimations() so should be able to do away with all of the followup calls in this test

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/harness.js
adreghiciu: Looks like if the test in not in there, it will not be run by maven build

kellyrob99: Damn :( 
We have been hopeful that we could keep both paths (mvn and browser) open to having all tests pass, which unfortunately is difficult without the ability to upload in the UI

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/search/search_basic.t.js
```diff
@@ -0,0 +1,152 @@
...
+                store = grid.getStore(),
+                model = store.getAt(0);
+
+            t.expect(model).toBeDefined();
+            t.expect(model.get('format')).toEqual('maven2');
+            t.expect(model.get('name')).toBe(t.anyStringLike('aopalliance'));
+            next();
+          }
+      )
+    }, 300000);
```
jdillon: This is the timeout stuff I mentioned before, why do we need this?  Shouldn't the default be sufficent?

kellyrob99: problem at the time was with BDD subtests, which with previous version at least seemed to expect that the timeout covered all of the encapsulated sub-cases. I can look again and see if it is still necessary - or if we're organizing something incorrectly.

adreghiciu:
CI +1 http://bamboo.s/browse/NX3-OSSF493-1
Added FT is passing

jdillon:
+1 we can/should sort out the timeout copy-pasta but doesn't have be tied to this PR.

kellyrob99:
+1

adreghiciu:
Manually merged https://github.com/sonatype/nexus-oss/commit/71723c576e2016f345e6a2c9d95b959474a43a71

PR #1496, opened 2015-06-30
Add NX.Assert and support for //<if assert> omission

jdillon:
related: https://github.com/sonatype/nexus-pro/pull/567

will log an error if boolean expression doesn't eval to true, not throwing exceptions as that messes up the entire app.

adds a omission flag to compile this out of the app.


djsauble:
+1

kellyrob99:
+1

adreghiciu:
+1

PR #1511, opened 2015-07-02
Make Message windowOptions available for internal customization

bradbeck:
Mostly so functional tests can do something like the following to control the timings associated with message windows.
```
  Ext.merge(t.controller('Message').windowOptions, {
    slideInDuration: 200,
    slideBackDuration: 200,
    autoCloseDelay: 300,
  });
```

### DIFF
--- a/plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Message.js
+++ b/plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Message.js
@@ -150,6 +150,17 @@ Ext.define('NX.controller.Message', {
   },
 
   /**
+   * Internal customization of message window options.
+   * At the moment, mainly intended for use by functional tests that need to
+   * override default settings.
+   *
+   * @private
+   * @property {Object}
+   */
+  windowOptions: {
+  },
+
+  /**
    * @public
    */
   addMessage: function (message) {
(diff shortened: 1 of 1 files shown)


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Message.js
```diff
@@ -150,6 +150,14 @@ Ext.define('NX.controller.Message', {
   },
 
   /**
+   * Internal customization of message window options.
+   *
+   * @private
```
jdillon: @property {Object}

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Message.js
```diff
@@ -150,6 +150,14 @@ Ext.define('NX.controller.Message', {
   },
 
   /**
+   * Internal customization of message window options.
```
jdillon: Should also say... used for functional-testsuite overrides of options ATM.

jdillon:
+1

adreghiciu:
+1

bradbeck:
http://bamboo.s/browse/NX3-OSSF510-1

There was an FT failure, but it is unrelated to this change.

PR #1522, opened 2015-07-03
NEXUS-8899 capabilities functional test

joedragons:
https://issues.sonatype.org/browse/NEXUS-8899
http://bamboo.s/browse/NX3-OSSF520

With pairing thanks to @DarthHater.
@kellyrob99 to manage merge if this is ready to move while I'm out.


testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/system/system_capabilities.t.js
```diff
@@ -0,0 +1,120 @@
...
+ * Tests system capabilities CRUD.
+ */
+StartTest(function(t) {
+  var waitForStoreToLoad = {
+    waitFor: 'storesToLoad', args: function() {
+        return t.cq1('nx-coreui-capability-list').getStore();
+    }
+  },
+      // Name of Capability tested
+      name = 'Base URL: Force';
```
mcculls: formatting is off here...

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/system/system_capabilities.t.js
```diff
@@ -0,0 +1,120 @@
...
+
+  t.describe('Given an admin when the admin is administering Capabilities', function(t) {
```
mcculls: Description doesn't follow the style used in other FTs, also could be shortened to "When an admin is administering Capabilities"

DarthHater: Easy enough. We were using a style more akin to Gherkin as a side note.

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/system/system_capabilities.t.js
```diff
@@ -0,0 +1,120 @@
...
+    t.it('then CRUD should work', function(t) {
```
mcculls: Other CRUD-based tests separate out each operation into different `it` methods (for example see https://github.com/sonatype/nexus-oss/blob/master/testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/repository/repository_crud.t.js#L51) which is imho easier to follow.

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/system/system_capabilities.t.js
```diff
@@ -0,0 +1,120 @@
...
+        // Verify that
+        function(next) {
+          var grid = t.cq1('nx-coreui-capability-list'),
+              store = grid.getStore(),
+              model = store.findRecord('typeName', name);
+
+          // we can find the updated record in store
+          t.ok(model, 'Capability ' + name + ' exists');
+          // and is it disabled
+          t.is(model.get('enabled'), 0);
```
mcculls: This assertion is currently failing on CI: http://bamboo.s/browse/NX3-OSSF520-TEST-1/test/case/153647346

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/system/system_capabilities.t.js
DarthHater: Agree with you, we were following the functional test for tasks which does not.  (https://github.com/sonatype/nexus-oss/blob/master/testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/system/system_tasks.t.js) it would be nice to clean this one up as it off from the pattern you describe.

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/system/system_capabilities.t.js
kellyrob99: should be true/false, not 0/1
```
is: No description
Failed assertion `is` at line 95 of testsuite-oss/system/system_capabilities.t.js
Got  : true
Need : 0
```

kellyrob99:
+1

mcculls:
+1

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/system/system_capabilities.t.js
joedragons: I realized this is dated and fixed and closed but I remember setting this
up because I thought each .it had to be able to run independantly and the
test Jeffry and I wrote update is not created by default so cannot be run
without create.
I checked Sencha and they say "All specs of the same nesting level are
executed sequentially." (ref:
http://www.bryntum.com/docs/siesta/#!/api/Siesta.Test.BDD-method-it)
Does that not say the same thing?

I am not sure where else I would have gleaned that (lots of info=)) but if
that's not true (or is just a style thing), then I agree it is easier to
follow and can easily do this style in the future.

Thanks for cleaning up after me(us), Kelly.
Peace,
Joe

mcculls: @joedragons the key thing is "All specs of the same nesting level are executed _sequentially_" so when you separate out the test into individual `it` blocks those are run one-by-one, and you can therefore rely on the earlier `it` block creating the capability running before the `it` block that updates the capability, with the `it` block that deletes the capability running last.

PR #1543, opened 2015-07-08
Fixes refresh, enables back by clicking feature menu

djsauble:
In user testing of milestone 4, a few users had difficulty going back after drilling into a view. They wanted to click the feature menu to go back, not the breadcrumb. This PR enables that behavior.

Before: http://take.ms/t70dK
After: http://take.ms/BWAvA

In addition, this PR fixes broken refresh behavior. Many of the views were refreshing some, but not all, of their stores (noted by @adreghiciu).

Before: http://take.ms/Xv2qS
After: http://take.ms/yb0Gv


plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Menu.js
```diff
@@ -110,9 +110,6 @@ Ext.define('NX.controller.Menu', {
         '#Permissions': {
           changed: me.refreshMenu
         },
-        '#State': {
-          changed: me.onStateChange
```
djsauble: @adreghiciu I removed a number of events which seemed to be redundant. For example, refreshing the app resulted in the 'featureselected' event being fired 13 times. By removing these events, I cut down on the redundancy, but since I don’t understand this code terribly well, it’s possible this has some negative effects. Let me know.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Menu.js
```diff
@@ -126,8 +123,7 @@ Ext.define('NX.controller.Menu', {
       },
       component: {
         'nx-feature-menu': {
-          select: me.onSelection,
-          afterrender: me.onAfterRender,
```
djsauble: @adreghiciu Another redundant event?

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Permissions.js
```diff
@@ -115,7 +115,6 @@ Ext.define('NX.controller.Permissions', {
     //</if>
 
     me.getStore('Permission').loadRawData(permissions, false);
-    me.firePermissionsChanged();
```
djsauble: @adreghiciu Another redundant event?

adreghiciu:
-1
* Why do we need to load the store in every controller when that can be done, as it was done a while ago in Drilldown?
* Removing sending of events/reacting on them may solve the menu reloading problem but for sure will break other parts. e.g. forms buttons will not longer enable/disable when permissions are changing, menus will not go away when status is changing (e.g. CLM is unconfigured, ...)
* Please split unrelated changes

djsauble:
Replaced by https://github.com/sonatype/nexus-oss/pull/1547 and https://github.com/sonatype/nexus-oss/pull/1549

PR #1544, opened 2015-07-08
Avoid health check calls to be buffered with other calls

adreghiciu:
because buffering will slowdown showing search results/browsed components until we get the health check results from data services


adreghiciu:
Closing in favour of https://github.com/sonatype/nexus-oss/pull/1545

PR #1547, opened 2015-07-08
Allows users to go back by clicking the feature menu

djsauble:
In user testing of milestone 4, a few users had difficulty going back after drilling into a view. They wanted to click the feature menu to go back, not the breadcrumb. This PR enables that behavior.

Before: http://take.ms/t70dK
After: http://take.ms/BWAvA


kellyrob99:
Seems to work as expected, but should also include an FT and corresponding CI build to help ensure no regressions

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/framework/go_back.t.js
```diff
@@ -0,0 +1,42 @@
...
+                testRepoRow = t.getRow(grid, grid.getStore().find('name', 'maven-public'));
+            t.click(testRepoRow, next);
+          },
+          function(next) {
+            t.waitForAnimations(next);
+          },
+          function(next) {
+            var tree = t.cq1('nx-feature-menu'),
+                selectedRow = tree.getSelectionModel().getSelection()[0];
+            tree.fireEvent('itemclick', tree, selectedRow);
```
kellyrob99: what about just t.click(selectedRow) here?

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/Menu.js
```diff
@@ -313,7 +311,9 @@ Ext.define('NX.controller.Menu', {
       if (node) {
         me.bookmarkingEnabled = me.navigateToFirstFeature;
         me.navigateToFirstFeature = false;
+        console.log(node.getPath('text'));
```
kellyrob99: cruft

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/framework/go_back.t.js
```diff
@@ -0,0 +1,37 @@
...
+ * of Sonatype, Inc. Apache Maven is a trademark of the Apache Software Foundation. M2eclipse is a trademark of the
+ * Eclipse Foundation. All other trademarks are the property of their respective owners.
+ */
+/**
+ * Test the ability to drill into a feature, then go back by clicking the feature menu
+ */
+StartTest(function(t) {
+
+  t.describe('Go back via the feature menu', function(t) {
+    t.it('Drill out', function(t) {
```
kellyrob99: this is not very descriptive, how about 'Can navigate back by clicking on menu item'

djsauble: Fixed. I figure we’ll use this FT to test the drilldown back button when that’s implemented, so I renamed things accordingly.

djsauble:
Feature test added, and passing locally. Ready for another round of review.

jdillon:
no objections, but some concerns on the failures bamboo reporting

kellyrob99:
@djsauble should be able to stabilize FTs with another merge from master branch, after which I am happy to help assess if any failures are regressions due to these changes

mcculls:
+1 verified new behaviour locally

kellyrob99:
Tests seem pretty unstable with these changes, and digging in I discovered that we're calling loadStores 4 times just logging into the page. Whether or not that is different from master I'm not sure, but changes here definitely seem to have affected the page load in a negative manner.
Log messages from the dev console: https://gist.github.com/kellyrob99/349170539fe8754527c5

mcculls:
@kellyrob99 it does feel slower when I compare it to master - is this related to the issue @adreghiciu mentioned during scrum where stores were being re-loaded even when they weren't active/visible?

kellyrob99:
@mcculls yes, it does seem so

kellyrob99:
Confirmed, on master Stores load much more sanely, even though we do seem to load the Repository store twice https://gist.github.com/kellyrob99/169a3ce325631f779388

djsauble:
@mcculls @kellyrob99 Thanks for taking a look at this. I’ll move this back to WIP.

djsauble:
I checked with @adreghiciu, and he said he implemented a fix for the store problem. I merged it in to this branch, and verified that duplicate store loading has stopped, but FTs are still failing.

:x: http://bamboo.s/browse/NX3-OSSF549-7

kellyrob99:
@djsauble How did you verify the store load? I'm still seeing multiple loads when navigating directly to the repositories page while not authenticated.

djsauble:
@kellyrob99 How can you navigate to the repositories page while not authenticated? This is what I see while authenticated.

repositories - sonatype nexus 2015-07-16 13-14-49

kellyrob99:
@djsauble http://localhost:8081/?debug#admin/repository/repositories when not signed in, get prompted for auth, after which I am still seeing 4 loads for each store.

kellyrob99:
+1 tests are passing so should be good to go back to master
http://bamboo.s/browse/NX3-OSSF549-15

jdillon:
Is this still a performance problem?

djsauble:
@jdillon The store loading issue has been resolved, so no, I don’t think there are any performance problems here.

PR #1559, opened 2015-07-10
Simplify setting of breadcrumb icon based on first column

adreghiciu:
Replace custom setting of breadcrumb icon with looking up the first column and using the icon class


djsauble:
Very nice. +1

kellyrob99:
+1, indeed good simplification here tnx @adreghiciu

bradbeck:
+1

PR #1566, opened 2015-07-13
NEXUS-8969 - Blobstore size calculations incorrect

kellyrob99:
- include content size from metadata in calculation

https://issues.sonatype.org/browse/NEXUS-8969
http://bamboo.s/browse/NX3-OSSF570-1

This is the first MapDB query I've written so please speak up if there are better patterns to use for iterating over all records with that framework.

PR #1570, opened 2015-07-14
NEXUS-8969 simple cache of calculated value

kellyrob99:
- based on number of entries only

enhancement to https://github.com/sonatype/nexus-oss/pull/1566


components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/MapdbBlobMetadataStore.java
```diff
@@ -109,7 +111,9 @@ protected void doStop() throws Exception {
   }
 
   private HTreeMap<BlobId, MetadataRecord> entries(final DB db) {
-    return db.getHashMap("entries");
+    return db.createHashMap("entries")
+        .counterEnable()
```
kellyrob99: this will have some impact on insert/update but otherwise ascertaining the size requires iterating over the entire collection

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/MapdbBlobMetadataStore.java
```diff
@@ -448,10 +452,17 @@ public long getBlobSize() {
     ensureStarted();
     return database.execute(new Fun.Function1<Long, DB>()
     {
-      @Override    
+      @Override
       public Long run(final DB db) {
         long totalSize = 0L;
         HTreeMap<BlobId, MetadataRecord> entries = entries(db);
+        if(entries.sizeLong() == sizeCache.getNumRecords()) {
```
jdillon: [anal] formatting

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/MapdbBlobMetadataStore.java
```diff
@@ -502,4 +514,36 @@ public Object invoke(final Object proxy, final Method method, final Object[] arg
...
+    }
+
+    public long getSize() {
+      return size;
+    }
+
+    /**
+     * Update the state of the cached data.
+     */
+    public void update(long numRecords, long size) {
```
jdillon: [anal] missing final on params assigned to same name of field

jdillon:
sure should hopefully avoid recalculating the size when it hasn't changed, and reduce i/o, though the metadata calc based on file-size could still be a wee problem of i/o

kellyrob99:
```
sure should hopefully avoid recalculating the size when it hasn't changed, and reduce i/o, though the metadata calc based on file-size could still be a wee problem of i/o 
```
@jdillon not sure what you mean here, size is stored in the metadata so we're not reading anything else from the filesystem during this calculation

jdillon:
@kellyrob99 getMetadataSize() is still presently calculated by inspection of java.io.File.length() is what I meant

```
  @Override
  public long getMetadataSize() {
    ensureStarted();

    // sum all file bytes in the database root
    long bytes = 0;
    for (File file : listFiles()) {
      bytes += file.length();
    }
    return bytes;
  }
```

kellyrob99:
@jdillon it is presented to the UI, but confined to just the metadata which is not expected to grow overly large - for 2000+ artifacts we have 3 files(not sure if that count will expand or not) at ~3MB total. Growth of this is linear based on number of artifacts so probably don't require optimization just yet.

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/MapdbBlobMetadataStore.java
```diff
@@ -64,6 +64,8 @@
 
   private TxMaker database;
 
+  private final SizeCache sizeCache = new SizeCache();
```
mcculls: I'd lean towards putting the `size` and `numRecords` fields here and avoid the cost of the extra lookup + method calls.

mcculls:
+1 but personally I'd drop the `SizeCache` abstraction since this is all internal

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/MapdbBlobMetadataStore.java
```diff
@@ -448,19 +460,28 @@ public long getBlobSize() {
     ensureStarted();
     return database.execute(new Fun.Function1<Long, DB>()
```
sonatype-zion: MAJOR Reduce this anonymous class number of lines from 25 to at most 20, or make it a named class. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1188)

sonatype-zion:
SonarQube analysis reported 1 issue:
* MAJOR 1 major

Watch the comments in this conversation to review them.

components/nexus-blobstore-file/src/main/java/org/sonatype/nexus/blobstore/file/MapdbBlobMetadataStore.java
jdillon: another rule to consider refining or disabling

kellyrob99: Aside from sonar then, that's a +1 ;)

jdillon:
did I already +1 this?

kellyrob99:
@jdillon no, but you did the linked issue, will merge now

PR #1584, opened 2015-07-20
Add intelligent defaults for blobstore selection

djsauble:
NOTE: Needs tests. However, I want to make sure this is the correct approach before writing those.

—

When creating a new repository…

1) If only one blobstore exists, select it
2) If there is more than one blobstore, select the one named 'default', if it exists.

Previous behavior was to leave the field blank, forcing extra user effort in all cases.

repositories - sonatype nexus 2015-07-20 10-54-27


plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/controller/Repositories.js
```diff
@@ -231,6 +234,27 @@ Ext.define('NX.coreui.controller.Repositories', {
 
   /**
    * @private
+   * When creating a new repository, select the ‘default’ blobstore by default (if it exists)
```
mrprescott: Comment doesn't match functionality

mrprescott:
+1

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/controller/Repositories.js
```diff
@@ -231,6 +234,27 @@ Ext.define('NX.coreui.controller.Repositories', {
...
+   */
+  selectDefaultBlobstore: function(combo) {
```
kellyrob99: Personally I think this should be a lot simpler - just select the first one in the list. 
And syntactically it should be as simple as:
```
combo.select(combo.getStore().getAt(0));
```

jdillon:
+0 not really worth spending too much time on this right now, added code to maintain may not be worth while given longer term issues we are likely to have related to marrying repositories to blobstores.

PR #1586, opened 2015-07-20
add method to wait for a spy to be called

kellyrob99:
as we're still seeing race conditions checking on message delivery

http://bamboo.s/browse/NX3-OSSF590-1


bradbeck:
+1

jdillon:
+1 sure

PR #1587, opened 2015-07-20
Disable save/discard buttons when form is invalid or not dirty

djsauble:
Before: http://take.ms/tptl5
After: http://take.ms/pIc3Q

http://bamboo.s/browse/NX3-OSSF591


jdillon:
Why does formBind not work?  IIUC that is the point of that field, why do we have to do special magic here?

djsauble:
@jdillon formBind only cares about the validity of a form, not its dirtiness.

Furthermore, if you try to let formBind handle the isValid() case, you can’t override the isDirty() case separately.

jdillon:
okay, thx for explanation @djsauble 

+1

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/SettingsForm.js
```diff
@@ -190,6 +195,20 @@ Ext.define('NX.view.SettingsForm', {
...
   /**
+   * Update the state of the save button
+   *
+   * @private
+   */
+  updateButtonState: function (form) {
+    var dirty = form.isDirty(),
+        valid = form.isValid();
+
+    if (form.down('button[action=save]')) {
```
kellyrob99: we should refrain from multiple executions of the same component query, extract a variable here please to avoid duplicate lookup and defining the actual CQ twice

kellyrob99:
Doesn't work if it's not tested ;)

djsauble:
@kellyrob99 What’s an adequate test? This could be as simple as:

1. Open Admin -> Repository
2. Choose a repository
3. Check whether the Save button is disabled

…or something much more complex:

1. Ensure that the save button is enabled when each type of form field becomes dirty
2. Ensure that the save button is disabled when each type of form field is reset
3. Ensure that the save button is enabled when each type of form field becomes invalid
4. Ensure that the save button is disabled when each type of form field becomes valid
5. Ensure that the save button is disabled after submitting a form

Is there a happy medium?

djsauble:
@jdillon Everything works as expected except for the Task Schedule form.

http://take.ms/T6Ihl

It’s a custom fieldset. I don’t understand why it’s not working correctly, but I’ll get that sorted prior to merge.

kellyrob99:
@djsauble The 100% correct answer would be to modify all existing tests to check the button state when forms are clean or invalid, but updating at least one that hits a settingsform would be a good start

testsuite/functional-testsuite/src/test/ft-resources/testsuite-oss/repository/repository_crud.t.js
```diff
@@ -121,6 +121,13 @@ StartTest(function(t) {
             t.is(form.down('#writePolicy').readOnly, false);
             t.is(form.down('#writePolicy').getValue(), 'ALLOW');
 
+            next(forms);
+          },
+          // confirm that the save button is disabled
```
djsauble: @kellyrob99 Something like this?

kellyrob99: no need to separate this into another function, but yes

djsauble:
FT added. Ready for another round of review.

I can’t reproduce the quirks I was seeing during demos. Either cache is to blame, or there’s some kind of race condition I can’t reproduce. Would appreciate more eyes on this.

joedragons:
I tested this generally (and generally looks good) and also was able to repro all the things I recalled from yesterdays demo so provided Daniel with info on them.  I'm not sure they're all bugs vs intentional so will let him sort it out.

joedragons:
@kellyrob99 If there are areas you know/see like this (or in general) where FT can be improved, feel free to keep me posted - I'm going to start keeping a list.  I think it'll be easier for me to improve the existing tests now that baselines are setup than create whole new ones.  Probably should/will touch base at some point=)

djsauble:
I’ve fixed the bugs that we noticed in demos yesterday. Ready for a second round of review.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/SettingsForm.js
```diff
@@ -125,6 +128,24 @@ Ext.define('NX.controller.SettingsForm', {
...
+  discardChanges: function (button) {
+    var form = button.up('form'),
+        record = form.getRecord();
+
+    if (form.api && form.api.load) {
+      form.fireEvent('load', form);
+    }
+    else {
+      form.getForm().reset();
+      form.isValid();
```
kellyrob99: what's the point of calling this please?

djsauble: Not sure I understand. This is the handler for the discard button on forms. It used to exist as a handler in the view, but I moved it to the controller as a listener (which is useful, so forms can define custom discard buttons and just reuse this logic).

repositories - sonatype nexus 2015-07-22 11-34-13

kellyrob99: Yes, but why call form.isValid() at all in this context?

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/SettingsForm.js
```diff
@@ -125,6 +128,24 @@ Ext.define('NX.controller.SettingsForm', {
   },
 
   /**
+   * Discard any changes to this form
+   *
+   * @private
+   */
+  discardChanges: function (button) {
+    var form = button.up('form'),
+        record = form.getRecord();
```
kellyrob99: not using this variable

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/SettingsForm.js
djsauble: This is @adreghiciu’s code. I bet we can remove that with no ill effects though.

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/support/SupportZip.js
```diff
@@ -125,8 +125,12 @@ Ext.define('NX.coreui.view.support.SupportZip', {
           text: NX.I18n.get('Support_SupportZip_Create_Button'),
           formBind: true,
           glyph: 'xf019@FontAwesome' /* fa-download */,
-          action: 'save',
+          action: 'add',
           ui: 'nx-primary'
+        },
+        {
+          text: NX.I18n.get('SettingsForm_Discard_Button'),
+          action: 'discard'
```
kellyrob99: I don't think this adds value on this form, more likely to be confusing since it's not clear what you're 'Discard'ing. Also, on a personal note I don't like that choice of wording and we might consider syncing up with CLM on this one - they use 'Cancel' for the same purpose.

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/SettingsForm.js
kellyrob99: aaah, OK.

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/support/SupportZip.js
```diff
@@ -125,8 +125,12 @@ Ext.define('NX.coreui.view.support.SupportZip', {
           text: NX.I18n.get('Support_SupportZip_Create_Button'),
           formBind: true,
           glyph: 'xf019@FontAwesome' /* fa-download */,
-          action: 'save',
+          action: 'add',
```
kellyrob99: action 'add' doesn't really fit with the use case

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/view/SettingsForm.js
```diff
@@ -187,6 +176,26 @@ Ext.define('NX.view.SettingsForm', {
   resetDirtyState: function (form) {
     var realform = form.getForm();
     realform.setValues(realform.getValues());
+    form.updateButtonState(form);
+  },
+
+  /**
+   * Update the state of the save button
```
kellyrob99: comment is not entirely correct

plugins/rapture/nexus-rapture-plugin/src/main/resources/static/rapture/NX/controller/SettingsForm.js
djsauble: Looks like a call to isValid() _might_ trigger client-side validation. I suppose that’s useful?

http://docs.sencha.com/extjs/4.2.3/#!/api/Ext.form.Basic-method-isValid

plugins/rapture/nexus-coreui-plugin/src/main/resources/static/rapture/NX/coreui/view/support/SupportZip.js
djsauble: ‘add’ is used by all the drilldown features when adding a new object. The behavior here is the same (since it should be possible to create a support zip when the form hasn’t been dirtied), but agree that the name is a bit misleading.

Rename all of the ‘add’ instances to ‘submit’?

kellyrob99: 'submit' makes sense to me

djsauble:
Passes FTs. Finally.

:white_check_mark: http://bamboo.s/browse/NX3-OSSF591-14

PR #1603, opened 2015-07-24
Make asset and component URLs transportable

djsauble:
Issue: https://issues.sonatype.org/browse/NEXUS-8838

Before: http://take.ms/rFMqp
After: http://take.ms/SbRB2

http://bamboo.s/browse/NX3-OSSF608-2


adreghiciu:
Tested the above: http://screencast.com/t/clF6vkSM

If the selected asset is not one from cached pages by infinite scroll it will not work.
There is also an error just by changing sorting
Entering a bookmark is not always resulting in navigation http://screencast.com/t/Xm50Ndobs

djsauble:
The scroll bug also exists in master. We should probably fix it separately. I’ve filed an issue here: https://issues.sonatype.org/browse/NEXUS-9016

Also, changing the bookmark now results in navigation. https://github.com/sonatype/nexus-oss/commit/3af8675561e57f0c90df90f165bd286bbcad125d

djsauble:
Talking with Kelly, we decided to create a separate issue for retrieving individual asset and component models from the server (https://issues.sonatype.org/browse/NEXUS-9024).

In the meantime, I’ve stubbed this PR to display a warning when we attempt to load a model which isn’t cached. https://github.com/sonatype/nexus-oss/commit/204f43634f470427053b9cdd0d138fbba85cdd07

This PR is ready for another round of review.

kellyrob99:
+1

PR #1605, opened 2015-07-24
Bind an instance of EhCacheManagerProvider for unit tests

bradbeck:


components/nexus-security/src/test/java/org/sonatype/nexus/security/AbstractSecurityTest.java
```diff
@@ -67,6 +70,15 @@ public final void doTearDown() throws Exception {
     tearDown();
   }
 
+  protected EhCacheManagerProvider createEhCacheManagerProvider() {
+    try {
+      return new EhCacheManagerProvider(getClass().getResource(ehCacheConfigPath).toURI());
+    }
+    catch (URISyntaxException e) {
+      throw new RuntimeException(e);
```
jdillon: throw Throwables.propagage(e) probably better.

components/nexus-security/src/test/java/org/sonatype/nexus/security/AbstractSecurityTest.java
```diff
@@ -67,6 +70,15 @@ public final void doTearDown() throws Exception {
     tearDown();
   }
 
+  protected EhCacheManagerProvider createEhCacheManagerProvider() {
```
jdillon: why protected?

components/nexus-security/src/test/java/org/sonatype/nexus/security/AbstractSecurityTest.java
```diff
@@ -47,6 +49,7 @@
 public abstract class AbstractSecurityTest
 {
   protected final TestUtil util = new TestUtil(this);
+  protected final String ehCacheConfigPath = "/ehcache-test.xml";
```
jdillon: why protected?

components/nexus-security/src/test/java/org/sonatype/nexus/security/AbstractSecurityTest.java
bradbeck: Ack.

components/nexus-security/src/test/java/org/sonatype/nexus/security/AbstractSecurityTest.java
bradbeck: In case the concrete unit test needs/wants to override the behavior.

components/nexus-security/src/test/java/org/sonatype/nexus/security/AbstractSecurityTest.java
bradbeck: In case the concrete unit test needs/wants to override the path to the configuration.

jdillon: any tests that presently need to do this?  if not I'd reduce visibility until its actually needed.

jdillon:
FTR I prefer this https://github.com/sonatype/nexus-oss/pull/1606 over the build complexity to pass around a config file.

Also FTR I was probably wrong originally when I removed the default if we couldn't find the configuration, so this is my bad for over simplifying the cache-manager component.  Didn't seem useful at the time, but I wasn't considering the IT that need a basic configuration to function and do not have a full assembly layout to properly find etc/ehcache.xml.

PR #1608, opened 2015-07-26
Ehcache 3

jdillon:
Experimentation to see if ehcache 3.0.0m1 is anywhere usable or not.

This kinda works, in UT but not in the server, freaks out with jaxb poops.


bradbeck:
Inadvertently closed.

jdillon:
Replaced by https://github.com/sonatype/nexus-oss/pull/1621

PR #1611, opened 2015-07-29
NEXUS-8980 migrate nexus.properties to org.sonatype.nexus.cfg

bradbeck:


mcculls:
+1

jdillon:
besides renaming the file, does this need to do something more to hookup/register with OSGI configadmin service thingy?

bradbeck:
As it is now. One can use the “config:…” commands to edit the properties from the karaf console. I discussed with @mcculls about adding a ManagedService implementation to intercept updates to properties: "the Jetty properties can't be changed without restarting the application, we could use a ManagedService to track the list of features to install but right now we don't need it so there's no pressing need to add such a service”

mcculls:
@jdillon adding a ManagedService is straightforward, the main question is what properties should be considered dynamic, ie. directly affect the running system. Looking at the current file the only one that springs to mind is the nexus-features property, where changes to that list would add/remove features - but even then there are other existing ways to add/remove features (though it would be interesting to persist changes made elsewhere back to the configuration file).

jdillon:
@mcculls no worries, just curious if we had to use the ConfigAdmin service instead of loading a properties file or something here.

`config:list` does show it:

```
----------------------------------------------------------------
Pid:            org.sonatype.nexus
BundleLocation: null
Properties:
   application-host = 0.0.0.0
   application-port = 8081
   felix.fileinstall.filename = file:/Users/jason/ws/sonatype/nexus/nexus-oss/target/nexus-base-template-3.0.0-SNAPSHOT/etc/org.sonatype.nexus.cfg
   nexus-context-path = /
   nexus-edition = nexus-oss-edition
   nexus-features = nexus-webresources-plugin,nexus-quartz-plugin,nexus-siesta-plugin,nexus-ssl-plugin,nexus-ldap-plugin,nexus-timeline-plugin,nexus-logging-plugin,nexus-atlas-plugin,nexus-wonderland-plugin,nexus-rapture-plugin,nexus-extdirect-plugin,nexus-coreui-plugin,nexus-repository-httpbridge,nexus-repository-nuget,nexus-repository-maven,nexus-repository-raw
   nexus-work = /Users/jason/ws/sonatype/nexus/nexus-oss/target/nexus-base-template-3.0.0-SNAPSHOT/data
   service.pid = org.sonatype.nexus
```

Unsure what BundleLocation is, or why some are null and some are refs, but that probably doesn't matter.

Anyways, +1

PR #1618, opened 2015-07-29
ConstraintViolationFactory

adreghiciu:
In limited situations we are not able to use jsr303 for validation, as for example in docker case where we need to check that configured port is not in use by other repository. In that case we cannot make a validator as the validator will need to know the repository the config belongs as otherwise it will find itself.

So, in such situations a non jsr303 validation is required, case when we will have to manually create ConstraintViolations to be able to throw ConstraintViolationExceptions.

This factory will do that, allowing to pass a bean path and a message.


jdillon:
FTR you could have created a helper bean with the information it needed, and a beanvalidation annotation for it, but thats likely overkill.

+1 to facility here to simplify, though we probably want to use this sparingly.

adreghiciu:
CI :+1: http://bamboo.s/browse/NX3-OSSF624-2

cstamas:
+1

PR #1623, opened 2015-08-07
[nexus-2.11.x] JRuby updates

mkristian:
first is use jruby-9k and duplicate more api under /api as more clients will use the new rubygems API.


dbradicich:
Closing in favor of newer impl in nx3

PR #1624, opened 2015-12-16
Add current user into http response header for logging purposes

flopma:
Hello,

Would it be possible to merge this code ?

The usecase is as follow. We have a ALM platform where we operates the full Atlassian software stack and Sonatype Nexus. All apps run on Tomcat and each runs behind a Apache Httpd instance. The latter allows us to log access and provide a ElasticSearch+Kibana for quick log queries.

So our Nexus runs on Tomcat perfectly and we have to modify the nexus-core JAR to include this code so we can get the authenticated user in Httpd access_log files.

Could you consider accepting this pull request ?

Thank you

### DIFF
--- a/components/nexus-core/src/main/java/org/sonatype/nexus/security/filter/authc/NexusHttpAuthenticationFilter.java
+++ b/components/nexus-core/src/main/java/org/sonatype/nexus/security/filter/authc/NexusHttpAuthenticationFilter.java
@@ -16,6 +16,7 @@ import java.io.IOException;
 import java.util.Map;
 
 import javax.inject.Inject;
+import javax.servlet.FilterChain;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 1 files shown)


peterlynch:
Nexus as a WAR  to run in other containers than Jetty is not a long term option we are focusing on.

https://support.sonatype.com/hc/en-us/articles/213465668-Where-is-the-Nexus-OSS-war-file-

Putting the requesting user name in the response headers in plain text could be considered by some customers as a security concern.

Instead of always editing the jar file, consider putting your custom class at `WEB-INF/classes/org/sonatype/nexus/security/filter/authc/NexusHttpAuthenticationFilter.class` where it will take precedence over the one in the jar file ( servlet spec behavior ).

Further the Jetty request.log lines in nexus 2.11.4 will always include the authenticated username. Consider running the full distributable and rely on the Jetty request.log instead.

flopma:
Thank you for your reply.

Actually, Atlassian products do it, putting authenticated username in response headers. And they are widely used in big enterprises.

And since we are talking HTTP applications, it is highly recommended to run them over SSL, so I don't see where the security concern is real.

Now, I know you guys are stopping the WAR delivery which we find a pity because you force customers to use the application server you chose. I explained our infrastructure, so you might understand that at operation level, it is easier to deal with one application servers than several, especially in our case where we have so many applications to manage.

Where are the days were Java applications where marketed as Servlet and JSP compliant and thus runneable on any application servers ?

Finally, thank you for the hint on the classes path. It will ease a little my deployment process.

dbradicich:
Closing, as noted above, it is highly recommended to use the app bundle which has the desired functionality

